PR #389, opened 2017-02-13
Basic HTML bookmark importer

broskoski:
screenshot 2017-02-12 21 40 26
screenshot 2017-02-12 21 41 02


broskoski:
Merging this so we can check it out on staging.

PR #424, opened 2017-03-25
remove block min-width

hxrts:
section extends beyond screen width on small iphone, removing min constraint


broskoski:
Thanks, Sam! This was actually fixed but not yet deployed (see: https://github.com/hxrts/ervell/blob/6699ae0c0fdfb849d3b0a7a6febf19324a831c66/apps/block/stylesheets/sidebar.styl#L13 ). Going to close this PR as we actually do need the min-width on larger sized screens.

hxrts:
perfect, thanks cab :)

PR #452, opened 2017-03-31
Deploy mobile changes and resultsCollection fix

broskoski:


dzucconi:
So this appears to have hidden the bookmarklet with a media query. I'm trying to piece together why.

Some contextualization here would have gone a long way.

apps/tools/stylesheets/tools.styl
```diff
@@ -161,4 +169,6 @@ li.friend-list__friend
 
 @media (max-width:800px)
   .tools__frame
-    width 100%
\ No newline at end of file
+    width 100%
+  #tab--bookmarklet
+    display none
```
dzucconi: This

components/path/stylesheets/tabs.styl
```diff
@@ -1,5 +1,7 @@
 .tab--container
   border-bottom 1px solid bordergray
+  +mobile()
+    display none
```
dzucconi: Oh actually, this is the rule.

PR #467, opened 2017-04-06
@dzucconi - redirect admins to index page, updates index styling

broskoski:
screenshot 2017-04-06 16 57 33


dzucconi:
Works for me

PR #488, opened 2017-04-24
Add support for inviting collaborator via email

broskoski:
collab


apps/channel/components/collaborators/new_collaborator_view.coffee
```diff
@@ -0,0 +1,82 @@
...
+    @render()
+
+  addCollaborator: (e) ->
+    @collaborators._add @results.get $(e.currentTarget).data 'id'
+
+  inviteCollaborator: (e) ->
+    @collaborators._invite $(e.currentTarget).data 'email'
+
+  isEmail: (query)->
+    is_email = new RegExp("[a-z0-9!#$%&'*+/=?^_`{|}~-]+(?:\.[a-z0-9!#$%&'*+/=?^_`{|}~-]+)*@(?:[a-z0-9](?:[a-z0-9-]*[a-z0-9])?\.)+[a-z0-9](?:[a-z0-9-]*[a-z0-9])?")
```
dzucconi: lol

apps/channel/components/collaborators/templates/results.jade
```diff
@@ -0,0 +1,10 @@
+include ../../../../../components/avatar/templates/avatar.jade
+
+for collaborator in collaborators
+  .new-collaborator__result.js-collaborator-result(data-id="#{collaborator.id}")
```
dzucconi: Try to avoid this unnecessary string interpolation:

```coffeescript
data-id="#{collaborator.id}"
# should be
data-id= collaborator.id
```

apps/channel/components/collaborators/templates/results.jade
```diff
@@ -0,0 +1,10 @@
...
+    span(class="iconic new-collaborator__result__add" data-glyph="plus" title="plus" aria-hidden="true")
+    .new-collaborator__result__name
+      | #{collaborator.get('username')}
```
dzucconi:
```jade
= collaborator.get('username')
```

apps/channel/components/collaborators/new_collaborator_view.coffee
```diff
@@ -0,0 +1,82 @@
...
+    'click .js-collaborator-invite' : 'inviteCollaborator'
+
+  initialize: ({ @channel, @collaborators} )->
+    @render()
+
+  addCollaborator: (e) ->
+    @collaborators._add @results.get $(e.currentTarget).data 'id'
+
+  inviteCollaborator: (e) ->
+    @collaborators._invite $(e.currentTarget).data 'email'
```
dzucconi: Why are all these methods `_private`. But then being used as if they weren't `_private`?

apps/channel/components/collaborators/new_collaborator_view.coffee
```diff
@@ -0,0 +1,82 @@
...
+
+module.exports = class NewCollaboratorView extends Backbone.View
+
+  events:
+    'keyup .new-collaborator__search' : 'searchCollaborators'
+    'click .js-collaborator-result' : 'addCollaborator'
+    'click .js-collaborator-invite' : 'inviteCollaborator'
+
+  initialize: ({ @channel, @collaborators} )->
+    @render()
```
dzucconi: I would try to avoid views that render inside of their initialize and just have the caller invoke it.

apps/channel/components/collaborators/new_collaborator_view.coffee
dzucconi: Probably want to extract this into a single utility function and use it here rather than it being a method on the view.

apps/channel/components/collaborators/new_collaborator_view.coffee
```diff
@@ -0,0 +1,82 @@
...
+    is_email = new RegExp("[a-z0-9!#$%&'*+/=?^_`{|}~-]+(?:\.[a-z0-9!#$%&'*+/=?^_`{|}~-]+)*@(?:[a-z0-9](?:[a-z0-9-]*[a-z0-9])?\.)+[a-z0-9](?:[a-z0-9-]*[a-z0-9])?")
+    is_email.test query
+
+  searchCollaborators: ->
+    query = @$input.val()
+
+    if @isEmail query
+      @showInvite query
+    else
+      $.get "#{sd.API_URL}/search/users/?q=#{query}&per=6", (response) =>
```
dzucconi: What happens when you rapidly type here? This method isn't throttled. Do these requests just start stacking?

apps/channel/components/collaborators/new_collaborator_view.coffee
```diff
@@ -0,0 +1,82 @@
...
+        @results = new Collaborators results
+        @showResults()
+
+  showInvite: (email) ->
+    @$('.js-collaborator-results').html inviteTemplate 
+      email: email
+
+  showResults: ->
+    @$('.js-collaborator-results').html resultsTemplate 
+      collaborators: @results.models
```
dzucconi: This should probably be a subView that listens to the results collection rather than something being manually called.

dzucconi:
I don't really agree with some of the view structuring, but I think this is generally fine.

The only real deal breaker is the unthrottled fetch. I'd throttle that and at least keep track of and cancel the pending requests.

apps/channel/components/collaborators/new_collaborator_view.coffee
broskoski: True, will do

apps/channel/components/collaborators/templates/results.jade
broskoski: yes, will change

apps/channel/components/collaborators/templates/results.jade
broskoski: yas

apps/channel/components/collaborators/new_collaborator_view.coffee
broskoski: I'm a creative soul.

apps/channel/components/collaborators/new_collaborator_view.coffee
broskoski: Will change

apps/channel/components/collaborators/new_collaborator_view.coffee
broskoski: Will throttle.

broskoski:
mergin!

PR #510, opened 2017-04-28
New profile test

broskoski:


components/block_v2/stylesheets/index.styl
```diff
@@ -0,0 +1,68 @@
+.Block
+  flex 1 1 315px
+  text-align center
+  margin 0 10px 90px 10px
+  max-width 315px
```
dzucconi: `315px` should be extracted and named something

broskoski: Yes will do this

components/block_v2/stylesheets/index.styl
```diff
@@ -0,0 +1,68 @@
...
+    
+    &--text
+      border 1px solid bordergray
+      display block
+      padding unit
+      overflow hidden
+      font-family font_serif 
+      text-align left
+      p
+        p()
```
dzucconi: Do these mixins actually correspond with out own font sizes and so forth defined in the variables file? (Of which there are already too many).

components/block_v2/templates/block.jade
```diff
@@ -0,0 +1,20 @@
+- extraClass = block.kind.__typename === 'Channel' ? "Block__inner--channel--" + block.kind.visibility : ''
```
dzucconi: I don't get this whole `extraClass` thing. Name it something meaningful or inline it or extract it to a helper.

dzucconi:
Lookin' goooood

broskoski:
Alrighty, when deployed this is just going to force admins to view this version of the profile over the current one.

PR #511, opened 2017-05-01
Check for element existence

dzucconi:
I'm more of the opinion that tabbing should be handled in the router and not with some media query show/hide. This is a bad pattern. Will fix when I redo the base tab component.


dzucconi:
This has no actual effect on the UI, just silences the error that Stripe is throwing.

PR #523, opened 2017-05-11
Init Connect V2

dzucconi:
The component doesn't do anything yet; just roughing in the look & feel of it.

connect

-----

Bonus round input sizes

### DIFF
--- a/apps/ui/client/index.coffee
+++ b/apps/ui/client/index.coffee
@@ -1,0 +1,4 @@
+Connect = require '../../../components/connect_v2/client/index.coffee'
+
+module.exports = ->
+  Connect $('.js-connect')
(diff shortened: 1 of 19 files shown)


apps/ui/client/index.coffee
```diff
@@ -0,0 +1,4 @@
+Connect = require '../../../components/connect_v2/client/index.coffee'
+
+module.exports = ->
+  Connect $('.js-connect')
```
broskoski: Noice

apps/ui/templates/pages/components.jade
```diff
@@ -0,0 +1,10 @@
+extends ../../../../components/layout/index--bare
+
+block body
+  .Page
+    .UI
+      .Connect.js-connect
+        include ../../../../components/connect_v2/templates/index
```
broskoski: Good idea

broskoski:
Looking beautiful as always

Issue #528, opened 2017-05-17
new connect QA


csherron:
block representation
- [x] add rounded corners to 'source' and connect buttons
- [x] add gray rounded edge border to block once connect interface opens
- [x] make 'close' button fixed to the bottom and 50px tall - fill empty space with more recent channels
- [x] make the 'create new private channel' title red
- [x] replace bottom result with the newly created channel once it gets created
- [x] connect block border only appears on hover state, leave it on the whole time

lightbox representation
- [x] center the 'connect' button
- [x] connect button hoverstate - instead of inverting colors, make medium gray normal, dark gray on hover
- [x] connect button padding: 20px top/bottom padding, 25px left/right padding
- [x] connect button margin - give like 20px more room above and below the button
- [x] connect button - use glyph instead of arrow icon for now

metadata representation
- [x] add arrow glyph next to 'Connect'

general
- [x] when text overflows, add hyphen and make sure it doesn't interfere with check mark, add a bit more space between lines  
- [x] is it supposed to be "Private Channel"? or "private channel?"
- [x] Lines break in Safari
- [x] recent channels list not persisting as long as before (newer public channels continue to show up)
- [x] on already connected channels, replace check mark w/ an 'x' on hover

list
- [x] fix styling

danpia:
- [ ] thicker line glitch on Chrome, disappears when you start typing
![unspecified-14](https://cloud.githubusercontent.com/assets/3718189/26167074/be4ba52e-3b03-11e7-855c-449db936ea3a.png)

csherron:
- [ ] on already connected channels, replace check mark w/ an 'x' on hover

csherron:
- [ ] connect block border only appears on hover state, leave it on the whole time

PR #531, opened 2017-05-18
Connect V2 Q/A

dzucconi:
Closes https://github.com/aredotna/ervell/issues/528

### DIFF
--- a/components/ui/buttons/overlay.styl
+++ b/components/ui/buttons/overlay.styl
@@ -1,5 +1,10 @@
 .Button
   &--overlay
+    // Fit 2-up on block hover
+    min-width (block_width / 2.25)
+    // Non-even vertical padding will visually center more accurately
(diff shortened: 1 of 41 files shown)


components/ui/buttons/overlay.styl
```diff
@@ -1,14 +1,23 @@
...
   &--overlay
+    // Fit 2-up on block hover
+    extra-padding = 0.125em
+    padding (button-v-padding + extra-padding) (button-h-padding + extra-padding)
+    min-width (block_width / 2.25)
+    font-size font-size-base
     background-color colors-gray-translucent
     border none
 
+    & + &
```
broskoski: woah whats this?

dzucconi: just the adjacent selector

dzucconi: like when `.foo` is next to another `.foo`, apply this style to the rightmost `.foo`

PR #537, opened 2017-05-19
Refactor connection list item

broskoski:
screenshot 2017-05-19 09 58 43
screenshot 2017-05-19 09 58 29


components/connection_list/index.jade
```diff
@@ -1,14 +1,12 @@
...
+  href=channel.href()
+  data-visibility=channel.getVisibility()
+) 
+  .Connection__inner
+    .Connection__inner__left
+      .Connection__inner__left__title= channel.get('title')
+      .Connection__inner__left__block-count &nbsp; #{channel.get('length')} blocks
+    .Connection__inner__right
+      | by&nbsp;
+      strong= channel.get('user').username
```
dzucconi: y u need so many divs :?

dzucconi: like the `inner` here can go completely

broskoski: Both sides need to truncate to ellipsis if they overflow

dzucconi: OK though you can do this and be a bit less vebose: similarly I did something on the item template for the connect menu: https://github.com/aredotna/ervell/blob/master/components/connect_v2/components/item/index.jade by just wrapping a span around things.

Like how about a top level span and then the rest of things follow normal BEM? Also get rid of the "left" and "right" designations, that's a pattern that then ties DOM structure to arbitrary styling decisions.

Also try to make these templates more legible so that you can look at them and see structure. I open a lot of files and see just a big blob of really long lines with no real visual indication of what's happening. Use line breaks.

components/connection_list/stylesheets/index.styl
```diff
@@ -0,0 +1,56 @@
+borderColor(color)
+  alpha(color, 30%)
```
dzucconi: what's this alpha helper do? `rgba(color, 0.3)` ?

components/connection_list/stylesheets/index.styl
```diff
@@ -0,0 +1,56 @@
...
+
+.Connection
+  height 45px
+  margin unit 0
+  padding (unit / 4) (unit / 2)
```
dzucconi: I'd start using ems for things where you can

components/connection_list/stylesheets/index.styl
```diff
@@ -0,0 +1,56 @@
+borderColor(color)
+  alpha(color, 30%)
+
+.Connection
+  height 45px
```
dzucconi: Don't hardcode heights

broskoski: What to do instead?

components/connection_list/stylesheets/index.styl
```diff
@@ -0,0 +1,56 @@
...
+      display table-cell
+      vertical-align middle
+      overflow hidden
+      text-overflow ellipsis
+      white-space nowrap
+    &__left
+      width 65%
+      overflow hidden
+      &__title
+        display inline
```
dzucconi: just use a span

components/connection_list/stylesheets/index.styl
dzucconi: padding + line height = whatever height it winds up being naturally.

components/connection_text_list/index.jade
```diff
@@ -0,0 +1,14 @@
...
+    else
+      if i == connections.length - 2
+        include _connection
+        | &nbsp;
+        span and&nbsp;
+      else if i == connections.length - 1
+        include _connection
+      else if i < connections.length - 2
+        include _connection
+        | ,&nbsp;
```
dzucconi: some comments might be in order?

components/connection_list/stylesheets/index.styl
dzucconi: let's just codify the concept of the border color and extract this to a higher level file

components/connection_text_list/index.jade
broskoski: Not changing this file, just updated the filename. Will address later.

components/connection_list/stylesheets/index.styl
broskoski: This opacity setting is specific to this connection list component at the moment.

components/connection_text_list/index.jade
dzucconi: I still don't know what this is lol

broskoski: Its the list of connections that shows up in the metadata of a channel

dzucconi: to my future self: it's `to_sentence`

dzucconi: fwiw some of these conditions don't appear to be needed:

```jade
.connection-list
  each connection, i in connections
    include _connection

    if i === connections.length - 2
      | &nbsp;
      span and&nbsp;
    else if i < connections.length - 2
      | ,&nbsp;
```

PR #538, opened 2017-05-19
Ordering of connections in Connect V2

dzucconi:
ordering-connections


dzucconi:
I guess we're gonna have some issues with the package now...

I might just rebase and kill that commit and then try to open a separate pull request for that. Let's see.

broskoski:
A+, nice specs

dzucconi:
Alright, sans package commits...

PR #540, opened 2017-05-19
Hide my buttons

dzucconi:


components/block_collection/stylesheets/block_collection.styl
```diff
@@ -382,6 +382,7 @@ body.is-mobile
   .grid__block__overlay
     opacity 1
```
dzucconi: Why opacity 1? No one knows.

broskoski: Someone does

Issue #530, opened 2017-05-17
New profile/blocks QA


danpia:
- [x] under Sort "Recently updated" should be black when applied
- [x] html not rendering under About
- [x] some images have Source button on hover

csherron:
New view modes

**All / Channels** (row view) **/ Blocks / Index** (index view)

row view
- [ ] missing "follow" and "start private channel" links under Info
- [ ] start rows 40px further down
- [ ] when browser is stretched out, container loses its center
- [ ] missing follow / connect hover on channels
- [ ] need throbber for loading new channel rows
 
index view
- [ ] remove focus thumbnails
- [ ] filter bar says "Filter Sort" should say "Filter X's Channels"
- [ ] channel borders make 0.3 opacity

PR #542, opened 2017-05-19
Various profile QA 

broskoski:
Closing out some more items in #530

### DIFF
--- a/apps/user/profile/index.coffee
+++ b/apps/user/profile/index.coffee
@@ -1,6 +1,9 @@
 Backbone = require 'backbone'
-{ QUERY, PROFILE_CHANNELS, SORT } = require("sharify").data
+{ QUERY, PROFILE_CHANNELS, SORT, USER } = require("sharify").data
 mediator = require '../../../lib/mediator.coffee'
+User = require "../../../models/user.coffee"
+PathView = require '../../../components/path/client/path_view.coffee'
+MetaEditableAttributeView = require '../../../components/editable_attribute/client/meta_editable_attribute_view.coffee'
 ChannelGroupView = require '../../../components/channel_block_group/view.coffee'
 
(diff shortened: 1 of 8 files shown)


apps/user/profile/index.coffee
```diff
@@ -55,7 +58,20 @@ class ProfileView extends Backbone.View
...
\ No newline at end of file
+  view.setUpChannelGroupViews(PROFILE_CHANNELS)
+
+  new PathView
+    el: $('section.path--header')
+    model: user
+
+  new MetaEditableAttributeView
+    model: user
+    el: $("#metadata--about .metadata__content")
```
dzucconi: From now on let's make it a point to only reference `js-` prefixed classes in javascript and to never style them.

broskoski: can do

components/connect_v2/integration/block_collection/view.coffee
```diff
@@ -24,7 +24,10 @@ module.exports = class BlockCollectionConnectIntegrationView extends Backbone.Vi
 
     this
 
-  remove: ->
+  remove: (e) ->
+    e.stopPropagation()
+    e.preventDefault()
```
dzucconi: I think there might not always be an `e` here which means this might fail. Either have another method that calls to it *or* use the existential operator

broskoski: Just curious, when would there not be an `e` here?

dzucconi: If a parent view invokes remove.

broskoski: Ah, got it.

broskoski:
Gonna merge this and continue!

PR #556, opened 2017-05-25
More profile QA

broskoski:


apps/user/templates/all.jade
```diff
@@ -0,0 +1,18 @@
+extends ../../../components/layout/index
+
+block body
+  include _path
+  if user && user.isPremium() && !following && !followers
+  .channel-container.channel-container--user( class=blocks.length ? '' : 'is-empty' )
```
dzucconi: This isn't indented under the condition?

Issue #571, opened 2017-05-30
Logged in navigation / Dropdown Q/A


dzucconi:
## All
- [x] Badge is perhaps too rounded?
- [x] Notifications unread alert style doesn't apply to channel title and should

## Desktop
- [x] ...

## Mobile
- [x] Notifications do not clear on mobile
- [x] Channel representation isn't square (but when it is square and full width it's too tall on 5s) - should just shrink the representation overall

----

## Related
- [x] Implement site-wide footer

PR #572, opened 2017-05-30
Dropdown Q/A

dzucconi:
Closes https://github.com/aredotna/ervell/issues/571


broskoski:
So, the footer will still show in all the places where there is infinite scroll?

dzucconi:
It will though it shouldn't. I'll amend but I'll have to do some modularization of these layouts.

PR #581, opened 2017-05-31
Update count name to fix index columns

broskoski:
This was colliding with another `count`, guessing notifications.

### DIFF
--- a/apps/user/templates/index.jade
+++ b/apps/user/templates/index.jade
@@ -7,8 +7,7 @@ block body
   include _path
   if _.keys(alpha).length
     .channel-container.flex-container.channel-container--alpha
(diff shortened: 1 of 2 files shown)


apps/user/templates/index.jade
```diff
@@ -7,8 +7,7 @@ block body
   include _path
   if _.keys(alpha).length
     .channel-container.flex-container.channel-container--alpha
-      - extraclass = count > 10 ? "alpha--two-columns" : ""
-      .alpha(class=extraclass)
+      .alpha(class=channelsCount > 10 ? "alpha--two-columns" : "")
```
dzucconi: There was a competing `count` in the scope too?

dzucconi:
Notifications are rendered client-side; but whatever works.

PR #584, opened 2017-05-31
Layout and footer

dzucconi:


dzucconi:
The whole layout component needs a bit more re-organization to make sense and I've left some `TODO`s but it's an OK start.

components/layout/index.jade
```diff
@@ -1,53 +1,41 @@
...
 
-  body(class=extraClasses)
+  body( class= extraClasses )
+    //- TODO: Refactor to use modalize, only thing that uses this is AuthView
     #modal-container
       //- Rendered client-side
 
-    #l-lightbox-container
-      //- Rendered client-side
-
```
dzucconi: I'm 99% sure nothing uses this. The only thing I can find is part of the block view ganks some styles from that component.

dzucconi: cc @broskoski

components/layout/index--infinite.jade
```diff
@@ -0,0 +1,41 @@
+- colorClass = sd.IS_INVERTED ? 'is-inverted' : ''
+- loggedInClass = user ? 'is-logged-in' : 'not-logged-in'
+- extraClasses = ''
```
dzucconi: I can't really refactor these out for now but whenever you include *any* variables inline within a template, ensure they are at the very top.

components/layout/index.jade
broskoski: Yeah, I think we can remove this.

broskoski:
So this will actually add the footer to the current setup, and you're going to add the "politeness" later?

dzucconi:
Infinite scroll stuff lacks footer. Everything else has it.

dzucconi:
"Politeness" is a much larger and more complicated issue and should be spec'd and opened separately.

broskoski:
👍

Issue #589, opened 2017-05-31
PDF blocks are missing a source button


csherron:

dzucconi:
When we fix this let's just refactoring the block representations so that we only have one of each kind. It's way too hard to keep track of all the different competing contextual implementations that have sprouted.

PR #610, opened 2017-06-01
Adds source to pdfs and mp3s

dzucconi:
Closes https://github.com/aredotna/ervell/issues/589

I realize if I fulfill my promise in that issue then we're gonna wind up with new blocks; so that's maybe a rabbit hole best saved for a bit later.

What is the difference between these two block templates @broskoski ?


broskoski:
Sigh, I think the block collection one still gets used in the feed sometimes.

PR #615, opened 2017-06-01
Existential check for attachment

broskoski:


dzucconi:
lol this template is brutal

PR #618, opened 2017-06-02
Marquee titles in connect component

dzucconi:


dzucconi:
linear-snap2

dzucconi:
Feels good

Issue #621, opened 2017-06-04
URL looses sort=random queryParam state when browsing back from block


damassi:
**To Repo:**
- Navigate to http://localhost:5000/explore?sort=random
- Click a Block
- Click the close button in the modal or hit back
- Notice that `/explore?sort=random` is replaced with `/explore`, resetting the users UI setting

PR #622, opened 2017-06-04
Bugfix: Preserve URLSearchParams on route change

damassi:
Fixes https://github.com/aredotna/ervell/issues/621

### DIFF
--- a/lib/setup.coffee
+++ b/lib/setup.coffee
@@ -81,6 +81,7 @@ module.exports = (app) ->
     app.use require("browserify-dev-middleware")
       src: path.resolve(__dirname, "../")
       transforms: [require("jadeify"), require('caching-coffeeify')]
+      debug: true
 
   # Test only
   if "test" is NODE_ENV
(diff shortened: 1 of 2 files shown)


lib/setup.coffee
```diff
@@ -81,6 +81,7 @@ module.exports = (app) ->
     app.use require("browserify-dev-middleware")
       src: path.resolve(__dirname, "../")
       transforms: [require("jadeify"), require('caching-coffeeify')]
+      debug: true
```
damassi: Adds better debugging / sourcemap support: 

Before:

screen shot 2017-06-03 at 7 48 21 pm

After:

screen shot 2017-06-03 at 7 47 49 pm

Maps to `--debug` in [Browserify](http://browserify.org/):
> Use --debug when creating bundles to have Browserify automatically include Source Maps for easy debugging.

broskoski: Thanks Chris!!

broskoski:
You are an 👼

Issue #612, opened 2017-06-01
Add Block Q/A


dzucconi:
- [x] https://github.com/aredotna/ervell/issues/529
- [x] https://github.com/aredotna/ervell/issues/536
- [x] https://github.com/aredotna/ervell/issues/208
- [x] Could do better job of rendering something usable from the server with a coherent disabled state then enabling once JS takes over
- [x] Button label should change when typing to note something like 'Add Block -> (or shift+enter)"
- [x] Adding should give some indication of a processing/loading state
- [x] Disable button when adding
- [x] Enable button when typing, disable when blank

PR #636, opened 2017-06-05
AddBlock refactor

dzucconi:
Closes https://github.com/aredotna/ervell/issues/612

add-block

add-block-list


dzucconi:
For some reason gifs wash out greys. But we'll peep it on staging.

apps/channel/client/block_skeleton_view.coffee
```diff
@@ -1,179 +0,0 @@
...
-    _.each @pages, (page)->
-      return false unless page.loaded
-
-    return true
-
-  cancelRequest: (id)->
-    @pages[id]?.xhr?.abort?()
-    @pages[id]?.loaded = false
-    @pages[id]?.loading = false
-    @unsetPlaceholdersLoading(id)
```
dzucconi: This didn't appear to be used anywhere.

apps/channel/client/index.coffee
```diff
@@ -0,0 +1,39 @@
...
+    blocks: blocks
+    blockCollectionView: view
+    resultsCollection: resultsCollection
+
+  initChannelPath channel
+
+  if current_user.canAddToChannel(channel) and not (FOLLOWERS? or FOLLOWING?)
+    addBlockView = AddBlockView $('.js-add-block'), blocks
+
+    resultsCollection?.on 'reset', ->
```
dzucconi: I don't entirely grok why this is needed but it seems to be.

apps/channel/client/block_skeleton_view.coffee
broskoski: Yes, you're right

dzucconi:
I forgot to delete the old new block code. For some good git line diff drama.

apps/channel/client/index.coffee
broskoski: When a user types in the filter and then erases, the resultsCollection needs to get cleared and the addBlock button needed to get re-initialized.

apps/statuses/index.coffee
```diff
@@ -0,0 +1,18 @@
+app = module.exports = require('express')()
+
+statuses = [
```
broskoski: What's this?

components/add_block/client/view.coffee
```diff
@@ -0,0 +1,100 @@
...
+      button: @$('.js-button')
+
+    @checkInput()
+
+  render: ->
+    @$el.html template()
+
+    @postRender()
+
+    this
```
broskoski: Much nicer 👌

apps/statuses/index.coffee
dzucconi: Just what I was developing against. So I can change the url to always fail or succeed and make sure states work as they should.

apps/channel/client/index.coffee
```diff
@@ -0,0 +1,39 @@
...
+  channelView = new ChannelView
+    el: $('.js-channel')
+    channel: channel
+    blocks: blocks
+    blockCollectionView: view
+    resultsCollection: resultsCollection
+
+  initChannelPath channel
+
+  if current_user.canAddToChannel(channel) and not (FOLLOWERS? or FOLLOWING?)
```
broskoski: Ok, this took a second to figure out, but there's going to be an error with this setup. 

The reason the new block view is getting initialized inside the actual `ChannelView` up there is this line:
```coffeescript
mediator.on 'collaborators:fetched', @checkUserAbilities, @
```
Channels where the user is a collaborator do not come down with every member of the channel, it has to be fetched after the fact.

broskoski: To clarify, this addBlock view will not show up now on channels where a user is a collaborator (and not the owner).

dzucconi: This is code that was copied from the existing implementation. But I'll read it now

apps/channel/client/index.coffee
```diff
@@ -0,0 +1,53 @@
...
+  initChannelPath channel
+
+  # Add Block initialization:
+  if current_user.canAddToChannel(channel) and not (FOLLOWERS? or FOLLOWING?)
+    integration = addBlockChannelServerRenderedIntegration
+      $el: $('.js-add-block')
+      collection: blocks
+
+    integration.init()
+    resultsCollection?.on 'reset', integration.reset
```
dzucconi: Default integration, now wrapped up into it's own thing.

apps/channel/client/index.coffee
```diff
@@ -0,0 +1,53 @@
...
+  mediator.once 'collaborators:fetched', (collaborators) ->
+    return if $('.js-add-block').length # Already rendered
+    return unless mediator.shared.current_user.id in collaborators.pluck 'id'
+
+    integration = addBlockChannelClientRenderedIntegration
+      $el: $('.js-block-collection')
+      collection: blocks
+
+    integration.init()
+    resultsCollection?.on 'reset', integration.reset
```
dzucconi: Client-side integration, similarly

dzucconi:
All should be well

broskoski:
👍

PR #644, opened 2017-06-07
Refactor find friends and add avatar

broskoski:
screenshot 2017-06-06 20 03 27
screenshot 2017-06-06 20 03 24


dzucconi:
Looks good to me

PR #658, opened 2017-06-08
Do not save if already saving

broskoski:
Closes https://github.com/aredotna/ervell/issues/655


dzucconi:
Two things:

1, Just use a simple variable for this. Disable and re-enable the button as well.

components/add_block/client/view.coffee
```diff
@@ -58,10 +61,13 @@ module.exports = class AddBlockView extends Backbone.View
     false
 
   save: ->
+    return false if @state.get('saving')
```
dzucconi: and you don't have to return false, just return

Issue #533, opened 2017-05-18
Pusher pushes `update` events on the connectable when a connection is removed from it


dzucconi:
This causes duplicate block representations to get pushed into whatever channel you happen to be in.

## To replicate

* Open the connect menu on a block in grid/list view
* Click connect to a different channel
* Then remove that connection
* A copy of that block will be pushed into the channel

PR #662, opened 2017-06-08
Do not trigger an add if the model is not found

broskoski:
This is just due to Backpusher making some stupid assumptions.  Sending an update when a connection is removed is technically correct, we just don't want to add the model if it doesn't already exist.

Closes https://github.com/aredotna/ervell/issues/533


dzucconi:
Maybe add a comment? It's a little weird that we're editing a vendored file like that.

broskoski:
Shall do

Issue #672, opened 2017-06-12
blocks / display QA


csherron:
- [x] unfollowing from the ‘following’ page, no feedback, users stick around even after refreshing
- [x] no way to connect channels in mobile
- [x] cant click on text blocks after they’ve been repositioned in a channel
- [x] "add block" button needs a hoverstate
- [x] follow button missing on channel blocks in mobile

broskoski:
What's `unfollowing from the ‘following’ page, no feedback, users stick around even after refreshing` mean? I can't replicate, can follow both users and the profile user in the following screen.

csherron:
go to your following section in your profile - unfollow someone, their profile block remains there, even after a refresh

dzucconi:
Assuming these are addressed from the checkmarks. Closing.

PR #677, opened 2017-06-12
Channel sort QA

broskoski:
Closes some items from https://github.com/aredotna/ervell/issues/672

### DIFF
--- a/components/add_block/stylesheets/grid.styl
+++ b/components/add_block/stylesheets/grid.styl
@@ -81,8 +81,6 @@
     pointer-events none
 
   &__button
-    &:disabled
-      pointer-events all
     &:hover
       color colors-gray-hover
     &:after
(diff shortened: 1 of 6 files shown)


components/add_block/stylesheets/grid.styl
```diff
@@ -81,6 +81,10 @@
     pointer-events none
 
   &__button
+    &:disabled
+      pointer-events all
```
dzucconi: Is this clickable then?

dzucconi: Because it should not be

dzucconi: And it shouldn't be hoverable anyway when it's disabled

broskoski: True, just a sec

broskoski: Confirmed that this rule is still applied after the user has typed

Issue #674, opened 2017-06-12
Mobile design tweaks, what happened to them?


csherron:
https://www.are.na/block/954001

https://www.are.na/block/954803

https://www.are.na/block/954005

https://www.are.na/block/953997

https://www.are.na/block/953996

dzucconi:
These existed?

csherron:
i remember cab showing them to me in staging a while ago

broskoski:
I will re-issue these and pr my work in progress branch

PR #678, opened 2017-06-12
[WIP] Fixed header for mobile

broskoski:
screenshot 2017-06-12 18 23 33

Start of #674


dzucconi:
What was the issue you were running into?

broskoski:
Looks like I had gotten through most of it, still needs to handle overflowing channel / user titles and updated styles for the mobile dropdowns.

PR #686, opened 2017-06-15
Add lib/reloadable for server-side code reloading

damassi:
@broskoski - Implemented this over in Force today and figured it'd be a good and easy Ervell win: https://github.com/artsy/force/pull/1552

I've also added an example app so you can test it out (http://localhost:5000/reload_app_example); be sure to remove.

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,6 +73,7 @@
     "browserify": "11.2.0",
     "browserify-dev-middleware": "*",
     "caching-coffeeify": "*",
+    "chokidar": "^1.7.0",
     "ezel-assets": "git://github.com/arenahq/ezel-assets.git",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -73,6 +73,7 @@
     "browserify": "11.2.0",
     "browserify-dev-middleware": "*",
     "caching-coffeeify": "*",
+    "chokidar": "^1.7.0",
```
damassi: Might need to rebuild `npm-shrinkwrap.json`

dzucconi:
Thank youuuuuu.

broskoski:
Thank you, dreamboat 🌃 ⛵️

PR #697, opened 2017-06-22
Cached homepage (demo blocks)

dzucconi:


dzucconi:
Noticed it's pretty slow uncached.

dzucconi:
Also I forgot to do this anyway

PR #705, opened 2017-06-26
Persist sort settings

broskoski:
This can be overridden by having a query param in the url, but the cookie only changes when you click the sort link. That make sense?

### DIFF
--- a/lib/middleware/sort.coffee
+++ b/lib/middleware/sort.coffee
@@ -1,0 +1,16 @@
+# Set sort from cookie
+#
+module.exports = (req, res, next) ->
+  # if the sort cookie is set, use that, otherwise default to updated_at
+  sort = if req.cookies['sort']? then req.cookies['sort'] else 'updated_at'
+  
+  # override the cookie if the request has a query param
+  sort = req.query.sort if req.query.sort
+  
+  # set the seed (for used in random sorting)
+  res.locals.sd.SEED = Math.floor(Math.random() * 100000000) + 1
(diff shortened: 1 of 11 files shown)


lib/middleware/sort.coffee
```diff
@@ -0,0 +1,16 @@
+# Set sort from cookie
+#
+module.exports = (req, res, next) ->
+  # if the sort cookie is set, use that, otherwise default to updated_at
+  sort = if req.cookies['sort']? then req.cookies['sort'] else 'updated_at'
```
dzucconi: `sort = req.cookies.sort or 'updated_at'` (unless sort is truthy)

models/state.coffee
```diff
@@ -6,13 +6,15 @@ module.exports = class State extends Backbone.Model
...
   initialize: ->
     # set values from cookies
     each keys(@defaults), (key) =>
       @set(key, val) if val = cookies.get key
 
     @on 'change:view_mode', @setCookie
+    @on 'change:sort', @setCookie
 
   setCookie: (model, value)->
     cookies.set keys(model.changed)[0], value
```
dzucconi: This is fine, but this model should probably be named something more specific like 'ViewState' or 'UIState'

broskoski: Can do this

lib/middleware/sort.coffee
broskoski: Tru

PR #715, opened 2017-06-27
Tools page clean up

dzucconi:
I'll circle back to this. I want to finish replacing the friend finder with the new one as well.


apps/tools/client/index.coffee
```diff
@@ -0,0 +1,17 @@
...
+    when 'find-friends'
+      model = new Authentication provider: 'twitter'
+      view = new FindFriendsView
+        el: $('#tab--find-friends')
+        model: model
+
+      model.fetch()
+
+    when 'send-invitation'
+      sendInvitation $('.js-send-invitation')
```
dzucconi: Replace with router

PR #719, opened 2017-06-27
[WIP] Handle home_path setting

broskoski:
This basically works, just needs to handle the custom home_path stuff. 

Makes me think though that maybe `/feed` should be the default setting on the attribute in the API?

### DIFF
--- a/apps/feed/index.coffee
+++ b/apps/feed/index.coffee
@@ -6,10 +6,18 @@ app = module.exports = express()
 app.set "views", __dirname + "/templates"
 app.set "view engine", "jade"
 
-app.get "/", routes.index
+homePathMiddleware = (req, res, next) -> 
+  return next() unless req.user
+  
+  path = req.user.get('home_path') 
+  return res.redirect 302, path unless path is "/" or !path
+  
+  next()
 
+app.get "/", homePathMiddleware, routes.index
+
+app.get "/feed", routes.index
 app.get "/notifications", routes.notifications
-
 app.get "/explore", routes.explore
 app.get "/explore/channels", (req, res, next) ->
   req.query = _.extend req.query, subject: 'channel'
--- a/apps/feed/index.coffee
+++ b/apps/feed/index.coffee
@@ -9,10 +9,13 @@ app.set "view engine", "jade"
 homePathMiddleware = (req, res, next) -> 
   return next() unless req.user
   
-  path = req.user.get('home_path') 
-  return res.redirect 302, path unless path is "/" or !path
+  path = req.user.get('home_path')
   
-  next()
+  # keep going in the middleware unless the path is "/" or the path is not present
+  return next() unless path or path is "/"
+
+  # otherwise redirect to the homepath
+  res.redirect 302, path
 
 app.get "/", homePathMiddleware, routes.index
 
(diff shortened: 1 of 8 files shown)


broskoski:
@dzucconi what say you

apps/feed/index.coffee
```diff
@@ -6,10 +6,18 @@ app = module.exports = express()
 app.set "views", __dirname + "/templates"
 app.set "view engine", "jade"
 
-app.get "/", routes.index
+homePathMiddleware = (req, res, next) -> 
+  return next() unless req.user
+  
+  path = req.user.get('home_path') 
+  return res.redirect 302, path unless path is "/" or !path
```
dzucconi: `!path` will never be evaluated, if it's `null` it's not going to be `'/'`

apps/settings/templates/tabs/general.jade
```diff
@@ -61,6 +61,26 @@ block content
...
+
+                option(
+                  value='/explore'
+                  selected= user.get('home_path') == '/explore'
+                ) Explore
+
+                option(
+                  value="/#{user.get('slug')}"
+                  selected= user.get('home_path') == "/#{user.get('slug')}"
+                ) Profile
```
dzucconi: How do you set a custom one?

lib/middleware/locals.coffee
```diff
@@ -10,6 +10,7 @@ _ = require 'underscore'
 module.exports = (req, res, next) ->
   res.locals.sd.CURRENT_PATH = parse(req.url).pathname
   res.locals._ = _
-  res.locals.homeHref = if parseInt(req.user?.get('following_count')) <= 1 then '/explore' else '/'
+  console.log('req.user?.homePath()', req.user?.homePath())
```
dzucconi: log

apps/settings/templates/tabs/general.jade
broskoski: Still need to handle that, hence the WIP

models/user.coffee
```diff
@@ -22,6 +22,9 @@ module.exports = class User extends Model
   isPremium: ->
     @get('is_premium')
 
+  homePath: ->
+    if parseInt(@get('following_count')) <= 1 then '/explore' else (@get('home_path') or '/')
```
dzucconi: OMG I never put it together that this was the behavior and was always confused.

dzucconi:
I think this approach with the middleware is 👌

apps/feed/index.coffee
dzucconi: Oh, I misread this. Still think it's a confusing condition.

broskoski: Yeah agreed, was trying to be extra careful to have it still evaluate even if the user doesn't have a `home_path` set. I think too careful to the point of being confusing.

dzucconi:
```coffeescript
return next() unless req.user.has('home_path')
```

instead of collapsing those different conditions

dzucconi: Does a req always have a user model, even if you're logged out?

broskoski: Good call. No, if you're logged out no user model will be present.

Issue #726, opened 2017-06-29
Missing analytics events


broskoski:
- [x] Upgrade to premium (`analytics.track.submit 'User paid for pro account'`)
- [x] Change default landing page

PR #728, opened 2017-06-30
Adds some analytics

broskoski:
Closes https://github.com/aredotna/ervell/issues/726


dzucconi:
I think it's disappointing that we're mixing the analytics code in with the real code... but that's where we're at, I guess.

broskoski:
@dzucconi what would you prefer?

dzucconi:
TBH I have no idea; there's no great solution. Separating it out at Artsy required a bit too much busy work for my taste, and we never fully separated the two.

Mixing it together sometimes lead to bugs; and generally just interferes with refactoring.

This is fine.

dzucconi:
I just hate analytics.

broskoski:
I know I know, but necessary

PR #766, opened 2017-07-08
Revert 763 revert 758 more analytics

dzucconi:
Re: https://github.com/dzucconi/ervell/commit/b5cb776cccda6838735c3210abcd2f19463cb6f9

The bookmarklet doesn't use the `eventBus` but needed to call trigger on it anyway. Thus err.


dzucconi:
🚌

PR #776, opened 2017-07-18
Make sure links get underlined in bolded text on the blog

broskoski:


apps/blog/stylesheets/show.styl
```diff
@@ -35,7 +35,7 @@
         font-size font-size-base
         line-height line-height-tall
 
-        > a
+        > a, > strong a
```
dzucconi: just make it a normal `a` selector

PR #781, opened 2017-07-19
[WIP] Use server to determine UI state for add block; others

dzucconi:


dzucconi:
Almost mergeable. Still hunting.

apps/channel/client/channel_view.coffee
```diff
@@ -41,14 +41,14 @@ module.exports = class ChannelView extends Backbone.View
...
       @setupFileDropView()
 
       @$('.js-block-collection').addClass 'is-addable'
       mediator.trigger 'channel:is-addable'
 
     # Editable
-    if collaborator or mediator.shared.current_user.canEditChannel(@channel)
+    if CAN.update
       @$('.js-block-collection').addClass 'is-editable'
       @$('.block-item').addClass 'can-manage'
```
dzucconi: I think I'll need to add some new abilities for the stuff below here. Still trying to make some sense of it.

apps/channel/routes.coffee
```diff
@@ -1,109 +1,102 @@
...
+  options = id: req.params.channel_slug
+  channel = new Channel options
+
+  canQuery = """
+    query channel_can($id: ID!) {
+      channel(id: $id) {
+        can {
+          add_to
+          update
+          follow: update
```
dzucconi: Probably makes sense to add this one as a distinct ability as well so we can guard against it on the server.

apps/channel/routes.coffee
```diff
@@ -1,109 +1,102 @@
...
-      channel: channel
-      blocks: blocks.models
-      author: author
-      followers: true
+      res.locals.sd.BLOCKS = channel.related().followers.toJSON()
+      res.locals.sd.FOLLOWERS = channel.related().followers.toJSON()
 
+      res.render 'index',
+        channel: channel
+        blocks: channel.related().followers.models # TODO: Odd naming
```
dzucconi: BLOCKS BLOCKS BLOCKS YOU'RE ALL BLOCKS

apps/channel/templates/_path.jade
```diff
@@ -18,7 +18,7 @@ block path
 
 block metadata
   .metadata.metadata--channel
-    if (user && author.id == user.id) || channel.has('metadata') || channel.get('collaboration') == true
```
dzucconi: True is True

components/blocks/container/templates/_grid_contents.jade
```diff
@@ -1,4 +1,6 @@
-if channel && !following && !followers && user && user.canAddToChannel(channel) && !sd.SHARE_TOKEN && !isSearching && !isEmbedded
+//- if channel && user && (can && can.add_to) && !following && !followers && !sd.SHARE_TOKEN && !isSearching && !isEmbedded
```
dzucconi: I thought I deleted you...

apps/channel/routes.coffee
broskoski: Hm, yeah good call. Update doesn't really correspond at all though currently.

apps/channel/routes.coffee
broskoski: :|

apps/channel/routes.coffee
dzucconi: That's what it was though

dzucconi: Oh, it's the opposite, rather.

apps/channel/templates/_path.jade
broskoski: If `(user && author.id == user.id)` is important though

dzucconi: I didnt change this. Look below. Ur missing the true is true

broskoski:
In general this looks great. Follow buttons will be broken though unless you reverse the condition.

apps/channel/templates/_path.jade
broskoski: Ah nevermind I see

apps/channel/routes.coffee
```diff
@@ -1,109 +1,102 @@
...
+      res.render 'index',
         channel: channel
-        blocks: blocks.models
-        author: author
+        blocks: channel.related().blocks.models
+        author: channel.related().author
+        can: can
 
-    .catch ->
+    .catch (e) ->
```
broskoski: This can be `.catch next`

dzucconi: Interestingly, it can't. Though the e is left over from debugging.

components/blocks/container/templates/_grid_contents.jade
broskoski: We still don't want an add block if the channel is an embed though...

dzucconi: And that's why there's no `can` on that route

dzucconi:
Alrighty, this feels good to go to me. Merging. Will deploy later on.

This wraps up the trickier aspects of switching the premium model. The rest is basically UI/X.

Issue #782, opened 2017-07-20
Any channel with less than 7 blocks reports an inaccurate view more


dzucconi:
### Steps to reproduce
* Create a channel and go to `/embed` or view one that has less than 7 blocks

### Expected result
Should not see a view more button

### Actual result
Displays a view more button with a negative count

![](http://static.damonzucconi.com/_capture/qvYw7qUOf1.png)

PR #785, opened 2017-07-24
Resets connection search after a connection is made

broskoski:
Also tried to fix https://github.com/aredotna/ervell/issues/782 for a while but couldn't figure out a way to do it with CSS only. I fear it will need more JS that gets bound on the window resize event.

### DIFF
--- a/components/connect/components/search/view.coffee
+++ b/components/connect/components/search/view.coffee
@@ -15,8 +15,14 @@ module.exports = class ConnectSearchView extends Backbone.View
     input: ->
       @query arguments...
 
-  initialize: ({ @state, @search }) ->
+  initialize: ({ @state, @search, @eventBus }) ->
     @query = throttle @perform, 200
+    
+    @eventBus.on 'connection:added', @afterConnection
+
+  afterConnection: =>
+    @$el.focus()
+    @$el.select()
 
   restore: ->
     @search.reset @collection.toJSON()
--- a/models/channel.coffee
+++ b/models/channel.coffee
@@ -23,6 +23,9 @@ module.exports = class Channel extends Block
   href: ->
     "/#{@get('user')?.slug}/#{@get('slug')}"
 
+  nMoreBlocks: (n) ->
+    @get('length') - n
+
   sync: (method, model, options) ->
     switch method
       when 'update', 'delete'
--- a/components/blocks/grid_item/stylesheets/embedded.styl
+++ b/components/blocks/grid_item/stylesheets/embedded.styl
@@ -12,14 +12,14 @@
     p:before
       content "+" attr(data-length) " "
     @media (min-width:(block_width_mar * 4) + 20)
-      p.lg
+      p.gt7
         display block
     @media (min-width:(block_width_mar * 3) + 20) and (max-width:(block_width_mar * 4) + 20)
-      p.md
+      p.gt5
         display block
     @media (min-width:(block_width_mar * 2) + 20) and (max-width:(block_width_mar * 3) + 20)
-      p.sm
+      p.gt3
         display block
     @media (min-width:0) and (max-width:(block_width_mar * 2) + 20)
-      p.xs
+      p.gt2
         display block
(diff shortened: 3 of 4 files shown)


models/channel.coffee
```diff
@@ -23,6 +23,9 @@ module.exports = class Channel extends Block
   href: ->
     "/#{@get('user')?.slug}/#{@get('slug')}"
 
+  nMoreBlocks: (n) ->
+    @get('length') - n
```
dzucconi: Can you make this method do the correct thing in the case of n being greater than the length?

components/connect/components/search/view.coffee
```diff
@@ -15,8 +15,14 @@ module.exports = class ConnectSearchView extends Backbone.View
     input: ->
       @query arguments...
 
-  initialize: ({ @state, @search }) ->
+  initialize: ({ @state, @search, @eventBus }) ->
     @query = throttle @perform, 200
+    
+    @eventBus.on 'connection:added', @afterConnection
```
dzucconi: `@listenTo`

models/channel.coffee
dzucconi: Tbh I don't really see the need for it anyway. It's a simple comparison in the template.

dzucconi:
I don't really understand why you'd want to be doing this with CSS. This is very weird to me which makes me think I'm missing something?

components/connect/components/search/view.coffee
```diff
@@ -15,8 +15,14 @@ module.exports = class ConnectSearchView extends Backbone.View
...
+
+  afterConnection: =>
+    @$el.focus()
+    @$el.select()
```
dzucconi: It's a view method so just use a normal arrow function.

dzucconi: I do think this is better than clearing the input, which would be weird IMO

broskoski:
Yeah, not as simple as you'd think. The embed shows a different number of blocks depending on the size of the window so that it's always an even number of blocks, with the last one being the see more block. We don't really want to remove that see more block if there are no more blocks in the channel, so it should turn into something like "view this channel". 

Apart from that, doing it in js means replicating the breakpoints again in js, which feels gross (and I think binding events on window resize always feels gross).

components/connect/components/search/view.coffee
broskoski: `on` isn't automatically bound to the view (was following your pattern in other views) but if listenTo works then cool.

dzucconi: > The advantage of using this form, instead of other.on(event, callback, object), is that listenTo allows the object to keep track of the events, and they can be removed all at once later on. The callback will always be called with object as context.

dzucconi:
I'm sure there's another way forward in terms of responsiveness but this is fine for now. However this doesn't actually fix the original bug. It just moved the incorrect logic into a model?

components/connect/components/search/view.coffee
broskoski: Yes, I know about listenTo, was following your pattern in this same component.

broskoski:
Right, why I said "tried" to fix.

components/connect/components/search/view.coffee
dzucconi: Where'd I use `on`? In something that doesn't extend Backbone.Events this is fine, and necessary.

broskoski: Yeah, looks like you used it out of the context of a view, makes sense https://github.com/aredotna/ervell/blob/master/components/connect/client/index.coffee

dzucconi:
Forgive me I'm not clever on Sunday nights.

```
remainder: (n) ->
  if (x = @get('length') - n) <= 0 then 0 else x
```

Then in the template just do a check on the remainder and if it's zero hide all the buttons?

dzucconi:
and obviously use that for the labels elsewhere.

broskoski:
Eh, +0 more blocks is not ideal but I guess better that what is there now. Would rather show "View this channel" or something similarly generic

dzucconi:
So just hide the damn thing if it's negative?

dzucconi:
If this really is complex I'll just look tomorrow.

broskoski:
Well, yeah I see what you mean. You'd have to do it for each variant I guess.

broskoski:
Lol it can't just hide, read my comments.

PR #789, opened 2017-07-24
Do not show tip if cookie is present

broskoski:
:/

### DIFF
--- a/apps/user/components/tips/index.coffee
+++ b/apps/user/components/tips/index.coffee
@@ -7,6 +7,7 @@ module.exports = ->
   tips = new Backbone.Collection TIPS
 
   tips.each (tip) ->
+    return if Cookies.get(tip.id) 
     new TipView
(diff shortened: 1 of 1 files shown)


apps/user/components/tips/index.coffee
```diff
@@ -7,6 +7,7 @@ module.exports = ->
   tips = new Backbone.Collection TIPS
 
   tips.each (tip) ->
+    return if Cookies.get(tip.id) 
     new TipView
       el: $(".js-tip[data-id='#{tip.id}']")
       model: tip
```
dzucconi: This one doesn't make sense to me. All this is doing is binding (or not) a view.

broskoski: Ah, yes you're right. Was probably running into a different issue that ended up showing a tip when it shouldn't. Closing.

Issue #805, opened 2017-07-27
Updated pricing page with private connection counter


dzucconi:
![](https://camo.githubusercontent.com/c3b76d748d6a241f0fbc4da028f21303ac86c866/687474703a2f2f7374617469632e64616d6f6e7a7563636f6e692e636f6d2f5f636170747572652f433633424c4e7a3777672e706e67)

PR #812, opened 2017-07-31
Implements simple connection count meter for non-premiums

dzucconi:
Closes https://github.com/aredotna/ervell/issues/805


dzucconi:
@broskoski note that I'm PRing against `premium_2` which will be the base branch for this milestone.

broskoski:
Got it, sounds good.

broskoski:
Nothing controversial here, proceed at will.

PR #820, opened 2017-08-03
[WIP] Supports coupons

dzucconi:
coupons

Still a bit more to do here, and I'm going to actually write some UI specs because this view is confusing. Just wanted to check in that this makes sense so far @danpia.


dzucconi:
Obviously the real limit is not 25, also

dzucconi:
coupons-cta

K, I'm gonna get this into the `premium_2` branch then pluck out what's needed.

dzucconi:
Actually I'll just wait for the copy stuff, while I'm here.

PR #822, opened 2017-08-03
Premium 2.0

dzucconi:
https://www.youtube.com/embed/MQgeRh0_yus

I don't think there's really anything here that can be deployed now.


dzucconi:
But I'll merge https://github.com/aredotna/ervell/pull/820 into this then we'll be in business.

PR #833, opened 2017-08-08
Keeps blocks count visible on long channel titles.

### LINKED ISSUE #774: Long channel titles in Index view prevent number of blocks from appearing
# Steps to reproduce

Go to Index view of most profiles.

screen shot 2017-07-14 at 2 32 54 pm

# Expected result

Number of blocks should always be visible (i.e. Channel title should be cut off earlier)

# Actual result

Long channel titles prevent the number of blocks from appearing. Eg. "A Local's Guide to Chicago Architecture" and "Architecture of the Atomic Age" in the Chicago Architecture Biennial profile.

asirgado:
Fixes #774.

### DIFF
--- a/components/connection_list/connection.jade
+++ b/components/connection_list/connection.jade
@@ -1,7 +1,10 @@
-span
-  .Connection__title
-    = channel.get('title')
-    span
+.Connection__item
+
+  .Connection__info
+    .Connection__title
+      = channel.get('title')
+
+    .Connection__blocks
       | &nbsp; #{channel.get('length')} blocks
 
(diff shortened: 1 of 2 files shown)


broskoski:
Hey @asirgado, thank you so much for these PR's. Unfortunately this issue is a little more complicated than it seems and we won't be able to merge this. When the `ConnectionItem` gets down to smaller sizes both the title and the author get truncated, which is not ideal. We will need to design something specific for these cases. 

Really appreciate your effort though!

7trmibiej5

PR #834, opened 2017-08-08
Adds support for video blocks (.mp4 & .mov files).

asirgado:
Hello Are.na team,

I was a little bit sad when I saw the attachment icon on video blocks:
screen shot 2017-08-08 at 13 30 37

I really want to create a channel for videos, especially now that VSCO is releasing video support! :)

From what I see on previous issues and code Are.na used to support it, no?

PS. There is no thumbnail attribute, that would make it easier for the grid view.

### DIFF
--- a/apps/block/stylesheets/index.styl
+++ b/apps/block/stylesheets/index.styl
@@ -124,9 +124,15 @@ block-sidebar-breakpoint = 950px
     fill textgray
 
 .block-video
+  margin 0 auto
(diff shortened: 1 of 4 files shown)


dzucconi:
While although I think this is a good idea it's really heavy performance/bandwidth-wise. The correct way for us to fix this would be on the backend: to extract thumbs from the uploads. Thanks anyway.

PR #840, opened 2017-08-09
Lined up headers

dzucconi:
Re:

We can't line things up with the landing page because that's vertically centered. Rest should be good.

### DIFF
--- a/apps/about/client/index.coffee
+++ b/apps/about/client/index.coffee
@@ -11,7 +11,15 @@ module.exports = ->
 
   scrollToId = (id) ->
     $target = $sections.filter("[id='#{id}']")
-    yPos = $target.offset().top
+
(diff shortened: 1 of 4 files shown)


apps/about/client/index.coffee
```diff
@@ -11,7 +11,15 @@ module.exports = ->
 
   scrollToId = (id) ->
     $target = $sections.filter("[id='#{id}']")
-    yPos = $target.offset().top
+
+    offset = if ($page = $('.js-page')).length
+      parseInt($page.css('margin-top'), 10) +
+      parseInt($page.css('padding-top'), 10)
+    else
+      0
```
dzucconi: Not my finest work.

broskoski: Welcome to my world

PR #845, opened 2017-08-09
Only set uiState cookies for logged in users

broskoski:


dzucconi:
To me this seems like the wrong level to handle this. (It's fine I don't really care)

broskoski:
Hm, where would you handle it?

dzucconi:
At the level of the caller: like why does a logged out user even need a UIState?

broskoski:
@dzucconi hows that sire

models/ui_state.coffee
```diff
@@ -1,6 +1,7 @@
 Backbone = require 'backbone'
 cookies = require 'cookies-js'
 { keys, each } = require 'underscore'
+CurrentUser = require './current_user.coffee'
```
dzucconi: not needed

components/path/client/path_view.coffee
```diff
@@ -17,7 +17,7 @@ module.exports = class PathView extends Backbone.View
     @setupSubViews()
 
   setupSubViews: ->
-    if (sd.USER or sd.CHANNEL) and sd.CURRENT_USER
+    if (sd.USER or sd.CHANNEL) and CurrentUser.orNull()
```
dzucconi: How's this any different?

PR #846, opened 2017-08-09
Check for user id on feed redirect

broskoski:


apps/feed/routes.coffee
```diff
@@ -3,7 +3,7 @@ sd = require('sharify').data
 ExploreBlocks = require '../../collections/explore_blocks'
 
 @index = (req, res, next) ->
-  return next() unless req.user
+  return next() unless req.user?.id
```
dzucconi: why can't we just null out the object on logout?

broskoski: `passport` is _supposed_ to handle that: https://github.com/aredotna/ervell/blob/master/apps/auth/routes.coffee#L27

PR #851, opened 2017-08-14
Grid refactor

dzucconi:
I'm opening and closing this so that I can take from it later (hopefully).

I set out to refactor the grid container on the channel view and encountered spaghetti the likes of which I have never before seen.

It's probably possible to get this to a mergeable state today but it feels pointless without just replacing the block representation completely.

Primary in my concern is how much we increased the surface for bugs by producing a poor copy of the block representation so that it could get it's data from GraphQL. There we had an opportunity to introduce a block that supported both and reduced the amount of spaghetti.

I feel there is no path to maintenance here and am cutting my losses until we have the time available to do that correctly.

### DIFF
--- a/components/pagination/skeleton_view.coffee
+++ b/components/pagination/skeleton_view.coffee
@@ -1,148 +1,148 @@
-Backbone = require "backbone"
 _ = require 'underscore'
-sd = require("sharify").data
-Blocks = require '../../collections/blocks.coffee'
+Backbone = require 'backbone'
 mediator = require '../../lib/mediator.coffee'
 
 module.exports = class SkeletonView extends Backbone.View
+  queue: []
+  pages: {}
+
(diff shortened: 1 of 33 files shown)


components/pagination/skeleton_view.coffee
```diff
@@ -1,148 +1,148 @@
...
 
-    return true
+    --@loading
 
-  cancelRequest: (id)->
+  cancelRequest: (id) ->
     @pages[id]?.xhr?.abort?()
     @pages[id]?.loaded = false
     @pages[id]?.loading = false
-    @unsetPlaceholdersLoading(id)
```
dzucconi: A lot of unneeded complexity here that I got rid of. This implements  the pagination without the need for those weird pagemarkers. This also removes the waypoints once they are used for a less janky scroll.

PR #869, opened 2017-08-22
Confirmable users, immediate registration

dzucconi:


apps/authentication/index.coffee
```diff
@@ -0,0 +1,23 @@
...
+app
+  .set 'views', "#{__dirname}/templates"
+  .set 'view engine', 'jade'
+
+app
+  # TODO: None of these paths make any kind of sense (snake_case;
+  # language: "sign up" vs "join"). Generate new routes and redirect to old.
+  .get '/sign_up', routes.sign_up
+  .get '/log_in', routes.log_in
+  .post loginPath, middleware.return # /me/sign_in
```
dzucconi: Maybe now is a good time to make these routes match the language we use. `sign_up => join`, `log_in => log-in`, etc...

apps/authentication/middleware.coffee
```diff
@@ -0,0 +1,37 @@
...
+@confirmable = (req, res, next) ->
+  request
+    .get "#{API_URL}/confirmations"
+    .set 'Accept', 'application/json'
+    .end (err, response) ->
+      return next(err) if err?
+
+      console.log response
+
+      res.json response
```
dzucconi: tbc

apps/authentication/templates/log_in.jade
```diff
@@ -0,0 +1,33 @@
...
+          required
+        )
+
+        a( href='/forgot' ) Forgot?
+
+    .Errors.js-errors
+      //-
+
+    button.Button.Button--divider.js-submit
+      | Log in
```
dzucconi: works without js

components/confirmable_message/index.styl
```diff
@@ -1,51 +1,31 @@
-.LoggedOutCTA
+.ConfirmableMessage
```
broskoski: 👍

apps/authentication/index.coffee
broskoski: Yeah agreed, it always confuses me + redirect old routes?

apps/authentication/templates/log_in.jade
broskoski: 👍

apps/registration/index.coffee
```diff
@@ -1,9 +0,0 @@
-express = require 'express'
-routes = require './routes'
-
-app = module.exports = express()
-
-app.set 'views', "#{__dirname}/templates"
-app.set 'view engine', 'jade'
-
-app.get '/register/:token', routes.acceptInvitation, routes.invalidInvitation
```
dzucconi: So invitations actually need to be rethought, now that I look at this.

I thought of putting it back, which I could still do: basically it'd just bypass confirmation. But then what happens if u want to sign up from a different email than what you've been invited?

I guess we'll have to put it back and then do a confirmation dance if you change it.

broskoski: Do we even allow people to change their emails after the fact?

dzucconi: Yeah. However before it was basically a security bug: you could sign up with one email then change it to an arbitrary one because we never "reconfirmed it" with this PR we enforce reconfirmations.

apps/util/routes.coffee
```diff
@@ -7,3 +8,12 @@ cache = require '../../lib/cache'
...
   res.redirect '/'
+
+@robots = (req, res) ->
+  res
+    .set 'Content-Type', 'text/plain'
+    .send switch NODE_ENV
+      when 'production'
+        'User-agent: *'
+      else
+        'User-agent: *\nNoindex: /'
```
broskoski: 👏

apps/authentication/routes.coffee
```diff
@@ -25,15 +26,16 @@ cache = require '../../lib/cache'
...
 
-        user = new User req.user.attributes
-        cache.del "#{user.url()}{}"
+        # IMPORTANT: return the `response` instead of the `user.toJSON()`
+        # Why? Because `user.toJSON()` is already parsed. Returning
+        # an already parsed response will make it unparesable.
+        res.json response
 
-        res.json req.user.attributes
+    .catch next
```
dzucconi: :sobs uncontrollably:

dzucconi: Hilariously, right now in production we do an account fetch: then if that's successful we do another one here on the refresh route!

PR #871, opened 2017-08-23
Nu-premium

dzucconi:
This is the sort of gist of what happens:

private

I can re-render the add block so that it switches to the CTA if y'all like, but that'll be a secondary PR. And if that's the case then I don't really see the point of this pop-up.

cc @danpia

### DIFF
--- a/apps/about/components/pricing_table/index.jade
+++ b/apps/about/components/pricing_table/index.jade
@@ -88,5 +88,5 @@
         | Access
 
       ul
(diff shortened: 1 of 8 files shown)


dzucconi:
Refined https://github.com/aredotna/api/pull/245 - ready to deploy that.

dzucconi:
(Obviously ignore the arbitrary limit I was testing with)

PR #876, opened 2017-08-28
Add @artsy/express-reloadable lib

damassi:
This is an iteration on https://github.com/aredotna/ervell/pull/686. The util has been pushed into its own package (https://github.com/artsy/express-reloadable) which removes all previous setup boilerplate.

Now, when `NODE_ENV=development` it mounts everything under apps and watches for changes, clearing nodes internal require cache on change (and page refresh). Changes under every app are now instantly available. Works much better than server restarts... cc @broskoski @dzucconi 

reloadable

One other thing I want to mention as a heads up (just because ervell and force are so similar) is https://github.com/artsy/stitch -- not sure how wedded you are to jade, but we're moving many of our sub apps over to React and this library popped out of it.


dzucconi:
thankyouuuuuuuuu

much nicer.

AND thanks for the tip re: stitch. We will be indeed looking at starting to migrate soon!

dzucconi:
Closing in favor of the rebased https://github.com/aredotna/ervell/pull/890

Issue #938, opened 2017-11-06
Onboarding Tweaks 03: Change profile and feed tips for new users


danpia:
## Profile

- [x] remove the current "empty state" sentence
- [x] change and restyle the current tips following the comp below
- [x] make the tips stick above the content until user closes them

![profile-tips](https://user-images.githubusercontent.com/3718189/32465595-07d68574-c312-11e7-997f-1c4200b1d509.png)

**Copy**
Welcome to Are.na! This is your profile. All your channels and connections will show up here.
	• 	Start adding content by creating your first channel. [opens channel creator]
	• 	Check out these favorite channels picked by the Are.na community. [link to https://www.are.na/are-na/classic-channels]

**UI**
_background: #f7f7f7, border-radius: 3px, font-size: 1.25rem font-size-lg_

## Feed/Explore
- [x] remove the current "empty state" sentence
- [x] add tips following the comp below
- [x] make the tips stick above the content until user closes them

![feed-tips](https://user-images.githubusercontent.com/3718189/32465770-aa72d260-c312-11e7-9ba6-c2b2384e4dd0.png)
**Copy**
This is your feed. Updates from the people and channels you follow will appear here.
	•	Go to Explore to see everything the Are.na community is connecting in real time. [link to explore]
	• 	Check out these favorite channels picked by the Are.na community. [link to https://www.are.na/are-na/classic-channels]

**UI**
_background: #f7f7f7, border-radius: 3px, font-size: 1.25rem font-size-lg_

broskoski:
@danpia doesn't seem like a great idea to remove a tip about installing the bookmarklet, what's the rationale there?

Issue #939, opened 2017-11-06
Onboarding Tweaks 04: corner pop-up tips for new users


danpia:
Stacked, from top to bottom:
1. "**Find friends from Twitter** on Are.na" [link to https://www.are.na/tools/find-friends ]
_show 1. only to people who haven't connected_
2. "Save to Are.na faster with **the bookmarklet**" [link to https://www.are.na/tools/bookmarklet ]
3. "If you need help on how to use Are.na, read our **FAQs** or contact us at help@are.na" [link to https://www.are.na/faqs]

broskoski:
@danpia I think we are gonna have to do something else with these, this seems a little nuts for a first time experience.
<img width="1395" alt="screenshot 2017-11-12 15 41 26" src="https://user-images.githubusercontent.com/821469/32703179-f87c5e66-c7bf-11e7-9dbe-8e8a5682c1da.png">

PR #948, opened 2017-11-10
Onboarding: Tips update

broskoski:
Closes https://github.com/aredotna/ervell/issues/938

screenshot 2017-11-10 18 05 32

Hold, going to do #939 here as well as they need to be deployed together.


broskoski:
screenshot 2017-11-13 10 52 32
screenshot 2017-11-13 10 52 14

Ok, this is ready to go, we decided to consolidate the two lower right messages back to the profile.

PR #960, opened 2017-11-28
Client-side Markdown fixes

dzucconi:
Doesn't really close https://github.com/aredotna/api/issues/398 but at least treats the symptom.


lib/markdown.coffee
```diff
@@ -0,0 +1,19 @@
...
+  breaks: true
+  pedantic: false
+  sanitize: true
+  smartLists: true
+  smartypants: true
+
+module.exports = (string) ->
+  string = string.replace /&gt;+/g, '>' # Support for blockquotes
+
+  marked(string, MARKDOWN_OPTIONS)
```
dzucconi: huge sigh

PR #965, opened 2017-11-29
DDeeppllooyy

dzucconi:


dzucconi:
TODOs on the API include

- Include entity rendering
- Adding a `minimal` option to the FormatMarkdown field in GQL for feed views.

PR #993, opened 2017-12-17
Add support for extended characters in channel index group headers

maxdeviant:
I created a channel with a name containing Chinese characters and noticed that it wasn't showing up in the channel index.

This PR adds support for extended characters (e.g., CJK, emoji) in the groupings on the channel index page.

I changed the channel grouping list to rely on JavaScript's built-in sorting rather than a whitelist.

A couple of things to note:

- [ ] The order of numbers and symbols has swapped; the symbol groups now come before the numbers
  - I can tweak this behavior so it stays the same, if it's a big deal
- [ ] Emojis don't appear to render correctly
  - If I `console.log(alphaKeys)` the emoji is already turned into '�'

See the screenshots below:

### Before

screen shot 2017-12-16 at 11 50 52 pm

### After

screen shot 2017-12-16 at 11 51 37 pm

screen shot 2017-12-16 at 11 52 02 pm

### DIFF
--- a/apps/user/templates/index.jade
+++ b/apps/user/templates/index.jade
@@ -1,16 +1,14 @@
 extends ../../../components/layout/index
 
 block body
-  - var range = '!~@#$%^&*()_-+=<>.,?/ABCDEFGHIJKLMNOPQRSTUVWXYZ'.split('')
-  - range.unshift('0-9')  
-
   include _path
   include _tips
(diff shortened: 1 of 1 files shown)


dzucconi:
Thannnnk you.

That emoji issue is curious but this is certainly an improvement.

Issue #986, opened 2017-12-05
CTA footer bar is missing on logged out channels


csherron:
_edited by cab_
- [ ] If the page is a channel, and the channel is open, change the copy to "Sign up for Are.na to add to [channel title]"
- [ ] if the page is a channel and the channel is closed, copy should be "Sign up for Are.na to follow [channel title]"
- [ ] if the page is a profile, copy should be "Sign up for Are.na to follow [username]"
- [ ] if none of these apply (private share link or block), the copy should be "Are.na is a platform for creative and collaborative research"

broskoski:
And if the page is a channel, and the channel is open, change the copy to "Sign up for Are.na to add to this channel"

PR #1012, opened 2018-01-03
Adds logged_out_cta

dzucconi:
Closes https://github.com/aredotna/ervell/issues/986

### DIFF
--- a/components/logged_out_cta/index.coffee
+++ b/components/logged_out_cta/index.coffee
@@ -1,0 +1,16 @@
+{ CURRENT_USER, USER, CURRENT_ACTION, CHANNEL } = require('sharify').data
+template = -> require('./index.jade') arguments...
+
+module.exports = ->
+  return if CURRENT_USER?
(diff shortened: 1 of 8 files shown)


broskoski:
Hold, this needs a spruce up. Sorry, should have added a screenshot of what it looked like before.

dzucconi:
This is what the confirm window looks like. So make it consistent.

broskoski:
screen shot 2018-01-03 at 5 23 23 pm

components/logged_out_cta/index.coffee
```diff
@@ -1,16 +1,23 @@
...
     when 'profile'
-      "Join Are.na to follow #{USER.full_name}"
+      "Join Are.na to follow #{truncate(USER.full_name, 35)}"
     else
       'Are.na is a platform for creative and collaborative research'
 
   $('body').append $el = $ template call: call
+
+  $el.one 'click', '.js-close', (e) ->
+    e.preventDefault()
```
broskoski: Set cookie here so if you close it once, it never comes back.

Issue #1011, opened 2018-01-03
explore on landing page


csherron:
<img width="828" alt="screen shot 2018-01-03 at 4 59 33 pm" src="https://user-images.githubusercontent.com/3780330/34541861-ecb35fce-f0a7-11e7-8be0-6696a76d10a3.png">

- [x] 480px between the carrot and the header
- [ ] 180px between the header and the blocks

<img width="863" alt="screen shot 2018-01-03 at 4 59 38 pm" src="https://user-images.githubusercontent.com/3780330/34541874-f71bf002-f0a7-11e7-84b2-1a42ba67935f.png">

- [x] 320px between the last row and "get started"
- [x] make "get started" bold, it's currently not bold right now

PR #1018, opened 2018-01-04
Explore on homepage

dzucconi:
Closes https://github.com/aredotna/ervell/issues/1011

So what's the params to filter by user @broskoski


broskoski:
@dzucconi check your slack

Issue #1020, opened 2018-01-04
Better error messages for GraphQL errors


broskoski:
This should be a 401 not a 500
![screen shot 2018-01-04 at 2 14 47 pm](https://user-images.githubusercontent.com/821469/34580098-afc99ba2-f159-11e7-801e-f9f0e654e8ec.png)

dzucconi:
graphql doesnt have status codes

dzucconi:
sorta

broskoski:
Regardless, this error page shouldn't look like this.

dzucconi:
ok assigning it to @csherron

broskoski:
@dzucconi it just needs to look like what it did before for now.

broskoski:
To reproduce, go to a private channel logged out

PR #1027, opened 2018-01-04
Convert the GraphQL error messages into some kind of matching status code

dzucconi:
Dear future self,

Sorry.

Closes https://github.com/aredotna/ervell/issues/1020

### DIFF
--- a/lib/middleware/error_status.coffee
+++ b/lib/middleware/error_status.coffee
@@ -1,0 +1,17 @@
+contains = (x, y) =>
+  x.toLowerCase().indexOf(y.toLowerCase()) >= 0
+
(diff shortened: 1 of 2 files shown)


lib/middleware/error_status.coffee
```diff
@@ -0,0 +1,15 @@
...
+
+  if contains(err.message, 'not found')
+    error.status = 404
+    error.message = 'Not found'
+
+  if contains(err.message, 'unauthorized') or contains(err.message, 'access denied')
+    error.status = 401
+    error.message = 'Access Denied'
+
+  next(error)
```
dzucconi: actually im gonna be a bit stricter here.

PR #1041, opened 2018-01-09
Correct logic for invite message

broskoski:


components/layout/client.coffee
```diff
@@ -160,7 +160,7 @@ initShortCuts = ->
 showInviteMessage = ->
   # Don't show this message if the current user is not logged in
   # or if they have a pending confirmation
-  return if CURRENT_USER?.is_pending_confirmation
+  return if !sd.CURRENT_USER or sd.CURRENT_USER?.is_pending_confirmation
```
dzucconi: Why would we want to show this if there is no logged in user

dzucconi: oh it's a return

dzucconi: This is correct.

PR #1042, opened 2018-01-09
Share refinements

dzucconi:


apps/channel/components/share/view.coffee
```diff
@@ -77,9 +82,11 @@ module.exports = class ChannelShareView extends Backbone.View
 
     options = map(properties, (v, k) -> "#{k}=#{v}").join ','
 
-    window.open $target.attr('href'), $target.data('service'), options
+    service = $target.data('service')
+
+    window.open $target.attr('href'), service, options
 
-    options
+    analytics.track.click 'Clicked a social share', service: service
```
broskoski:
```coffeescript
analytics.track.click 'Clicked a social share'
  label: "Service"
  value: service
```

https://github.com/aredotna/ervell/blob/master/lib/analytics.coffee#L65

dzucconi: you can't pass more than one key/value to goog?

Issue #1026, opened 2018-01-04
add 'share' link to channel descriptions and light box 


csherron:
updated below

csherron:
!reminder not to include share links on private channels / private blocks

csherron:
updated

click share then 3 options appear

1. Copy link (or "private" link if a private channel (private block?))
2. Twitter
3. Facebook

clicking "copy" still shows orange feedback text 
clicking a "copy private link"  reveals the private share link
![screen shot 2018-01-05 at 1 20 17 pm](https://user-images.githubusercontent.com/3780330/34622965-43045ca4-f21d-11e7-9136-5b6d11375002.png)

![screen shot 2018-01-05 at 1 20 22 pm](https://user-images.githubusercontent.com/3780330/34622974-4ce18594-f21d-11e7-87bc-3a28f555b446.png)
![screen shot 2018-01-05 at 1 22 36 pm](https://user-images.githubusercontent.com/3780330/34622977-50e35a1e-f21d-11e7-9d9c-6cfb10589363.png)
![screen shot 2018-01-05 at 1 26 57 pm](https://user-images.githubusercontent.com/3780330/34622981-52e9cb86-f21d-11e7-8dd7-09733ee55072.png)

dzucconi:
- [ ] always display info column in channel

PR #1044, opened 2018-01-09
Block share

dzucconi:
Closes https://github.com/aredotna/ervell/issues/1026

### DIFF
--- a/apps/block/components/share/index.coffee
+++ b/apps/block/components/share/index.coffee
@@ -4,9 +4,7 @@ BlockShareView = require './view.coffee'
 generateText = ({ title }) ->
   truncate(title, 50)
 
-module.exports = ({ block }) ->
-  $el = $('.js-block-share')
-
+module.exports = ({ $el, block }) ->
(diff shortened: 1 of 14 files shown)


apps/block/components/share/index.coffee
```diff
@@ -0,0 +1,15 @@
...
+
+module.exports = ({ block }) ->
+  $el = $('.js-block-share')
+
+  view = new BlockShareView
+    text: generateText(block.toJSON())
+
+  $el.html view.render().$el
+
+  view
```
dzucconi: Woops lemme make sure this gets cleaned up in block views.

dzucconi:
Good lord this is ugly.

apps/block/components/share/index.coffee
broskoski: Cleaned up as in torn down when BlockView gets removed?

dzucconi: Yes. which now it does.

PR #1068, opened 2018-01-23
Removes ubuntu install; uses Node 8

dzucconi:


dzucconi:
I'm just trying to confirm if we need that step

PR #1071, opened 2018-01-24
Various minor fixes

broskoski:


apps/channel/templates/_path.jade
```diff
@@ -25,9 +25,11 @@ block metadata
 
     include meta_pockets/connections
 
-    if user
+    if can.update
       include meta_pockets/privacy
       include meta_pockets/export
+
+    if can.manage
```
dzucconi: woof

dzucconi: I'd rather u say `if can.destroy`

broskoski: why woof

broskoski: Can do

dzucconi: Not that the server would have let anyone delete anything... It just must have been disconcerting to see that button when you shouldn't.

PR #1094, opened 2018-02-06
Webpack

dzucconi:
Still some to clean up here.

I basically have deploys working; some stuff to sort out with environments.

Thanks to @damassi for a large amount of copypasta.

I'll ask you @broskoski to check out this branch and get it running locally once I'm through.


components/layout/client.coffee
```diff
@@ -34,6 +35,28 @@ module.exports = ->
...
   initConfirmableMessage()
   initLoggedOutCTA()
+  initGlobalBlockRouting()
+
+initGlobalBlockRouting = ->
+  # TODO: Extract and init block router only
+  # where we actually need it.
+  # Since migrating to Webpack, histories can conflict.
+  # Simply starting one history isn't working as it should,
+  # due to split packages (I believe).
```
dzucconi: This is annoying. On the one hand, we shouldn't have this problem to begin with. On the other I can squint and imagine situations in which you might need two routers.

damassi:
This makes me happy 😄 Let me know if you run into any weirdness

dzucconi:
Alright now just to get Travis deploys in order.

.babelrc
```diff
@@ -0,0 +1,17 @@
...
+      "targets": {
+        "browsers": ["last 2 versions"]
+      }
+    }],
+    "stage-3"
+  ],
+  "plugins": [
+    "transform-runtime",
+    ["module-resolver", {
+      "root": ["./"]
```
damassi: Note that in JS files this plugin allows for absolute imports, so no more `../../../` unmaintainability:
```javascript
import artworksApp from 'desktop/apps/artworks'
```

dzucconi: Ah, yes of course! Game changer!

dzucconi:
I haven't enabled hot reloading yet. Maybe once I get something in React working. It always seems like that stuff never works/more trouble than it's worth. But we'll see.

damassi:
For our react stuff in force (and positron for a more contained example) hotreloading works really well. The key is [this bit](https://github.com/aredotna/ervell/pull/1094/files#diff-11e9f7f953edc64ba14b0cc350ae7b9dR48) which will automatically detect if you're editing react code and hot swap out, and if you're in old coffeescript / not-react code[this bit](https://github.com/aredotna/ervell/pull/1094/files#diff-11e9f7f953edc64ba14b0cc350ae7b9dR174) will detect and automatically reload the whole page, so no more `cmd+r` which is really nice.

.babelrc
damassi: If only it worked in coffeescript files :/

dzucconi:
Yah, I'll definitely give the hmr a go in a moment while I try to write a feature in React now. Thanks again @damassi for paving the way here.

PR #1096, opened 2018-02-07
Configurable worker amount

dzucconi:


package.json
```diff
@@ -98,17 +98,13 @@
   },
   "devDependencies": {
     "benv": "*",
```
dzucconi: They don't actually run but we actually have specs in this app that use benv. I need to do a pass on all the testing setup and come up with a path forward.

PR #1100, opened 2018-02-08
Gets React/Apollo working

dzucconi:


apps/graphql/index.js
```diff
@@ -0,0 +1,31 @@
+import axios from 'axios';
+import express from 'express';
+import sharify from 'sharify';
+
+import { NODE_ENV } from 'config.coffee';
+
+const app = module.exports = express();
+
+const { data: { GRAPHQL_ENDPOINT, X_APP_TOKEN, CURRENT_USER } } = sharify;
```
dzucconi: I actually don't understand how sharify is working with setting and unsetting current user on a per-request basis if you're able to require it like this?

dzucconi: Actually it is probably just the case that this is null here. Of course.

PR #1101, opened 2018-02-08
Gets React/Apollo working

dzucconi:


apps/graphql/index.js
```diff
@@ -0,0 +1,29 @@
...
+
+const app = module.exports = express();
+
+app.post('/graphql', (req, res, next) => {
+  if (NODE_ENV !== 'development' && req.hostname !== 'are.na') {
+    const err = new Error('Requests must originate from are.na');
+    return next(err);
+  }
+
+  const X_AUTH_TOKEN = req.user && req.user.get('authentication_token') || '';
```
dzucconi: This only works if we're on the server.

lib/apollo/index.js
```diff
@@ -0,0 +1,38 @@
...
+import { ApolloClient } from 'apollo-client';
+import { createHttpLink } from 'apollo-link-http';
+import { setContext } from 'apollo-link-context';
+import { InMemoryCache } from 'apollo-cache-inmemory';
+
+const httpLink = createHttpLink({ uri: '/graphql' });
+
+const authLink = setContext((_, { headers }) => {
+  const { data: { CURRENT_USER } } = sharify;
+  const X_AUTH_TOKEN = CURRENT_USER && CURRENT_USER.authentication_token || '';
```
dzucconi: And this only works if we're on the client.

dzucconi:
react

apps/graphql/index.js
```diff
@@ -0,0 +1,29 @@
+import axios from 'axios';
+import express from 'express';
+
+import { NODE_ENV, GRAPHQL_ENDPOINT, X_APP_TOKEN } from 'config.coffee';
+
+const app = module.exports = express();
+
+app.post('/graphql', (req, res, next) => {
+  if (NODE_ENV !== 'development' && req.hostname !== 'are.na') {
+    const err = new Error('Requests must originate from are.na');
```
broskoski: 👍

PR #1104, opened 2018-02-13
Remove fastclick

broskoski:
This is breaking things in production right now. Let's live without it for the time being.


dzucconi:
What does breaking things mean.

dzucconi:
Ah well yep it is.

PR #1107, opened 2018-02-16
Re-implementation of collaborator management in React

dzucconi:
This does not add new functionality but rather replaces the existing collaborators implementation with one in React.


react/components/CollaboratorsList/index.js
```diff
@@ -0,0 +1,54 @@
...
+const StyledCollaboratorsList = styled(CollaboratorsList)`
+  margin-bottom: 1em;
+`;
+
+class CollaboratorsListContainer extends Component {
+  static propTypes = {
+    channel_id: PropTypes.number.isRequired,
+    // HACK: Until we wire up SSR; this requires an HTML fragment
+    // to render while loading.
+    htmlFragment: PropTypes.string.isRequired,
```
dzucconi: This is probably the only mildly controversial thing in this PR. I haven't wired up SSR yet so instead this just takes over the HTML fragment while it's loading so we don't get a blank flash.

react/components/Avatar/fragments/avatar.js
```diff
@@ -0,0 +1,9 @@
+import gql from 'graphql-tag';
+
+export default gql`
+  fragment Avatar on UserInterface {
+    href
+    initials
+    avatar(size: LARGE)
+  }
+`;
```
broskoski: Wow, you really did reinvent the world. I commend you on these efforts.

apps/channel/client/index.coffee
```diff
@@ -28,12 +29,14 @@ module.exports = ->
...
-  initManageCollaborators
-    collection: collaborators
-    current_user: current_user
-  initCollaboratorsList
-    collection: collaborators
+
+  $collaboratorsListEl = $('.js-collaborators-list')
+  mount CollaboratorsList, {
+    channel_id: channel.id
+    htmlFragment: $collaboratorsListEl.html(),
```
broskoski: Ah I see, I think this is a fine compromise for now if you do.

react/components/UI/Modal/index.js
```diff
@@ -0,0 +1,38 @@
...
+    this.el = document.createElement('div');
+  }
+
+  open = () => {
+    document.body.appendChild(this.el);
+
+    this.Provided = provide(this.Component, {
+      onClose: this.close,
+      ...this.props,
+    });
```
dzucconi: This is slightly weird. I might re-implement this using portals. Or have it be an option. The modal winds up being in a separate React DOM tree. I figured we might want to build multiple UIs as modals that are distinct while migrating over. Since this is client-side they wind up sharing the same Apollo store anyway.

apps/channel/client/index.coffee
dzucconi: Yeah, I'll get to it. The benefit is just being able to delete the old code. We'll be using @artsy/stitch here. Todos, todos.

react/components/UI/Modal/index.js
broskoski: Hm yeah. I'll add that it took a second for me to parse `provide` as terminology (env though I love the idea). Maybe something more descriptive like `wrapApollo` is more quickly understandable.

PR #1130, opened 2018-03-08
Update Travis API key

broskoski:


dzucconi:
What changed here?

broskoski:
I think the API key needed to be regenerated after setting up two-factor

PR #1133, opened 2018-03-14
Add VSCode cmd+click to symbols / eslint fix on save

damassi:
For cmd+click / cmd+hover over symbols - also works in JSX: 

click-to-def

(Note that if you avoid `export default` and use named exports only you get better static analysis for things like this, which will jump right to the symbol rather than to the top of the file)

And some `@types` for intellisense: 

screen shot 2018-03-13 at 10 11 13 pm

Also turned on eslint autofix-on-save: 

fix-on-save

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,4 @@
 // Place your settings in this file to overwrite default and user settings.
 {
+  "eslint.autoFixOnSave": true
 }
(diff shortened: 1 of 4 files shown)


dzucconi:
oh wow that click to file!

dzucconi:
thanks!

PR #1141, opened 2018-03-16
Minor styling fixes

### LINKED ISSUE #1127: collaborator modal top gets cut off on my 13" screen
screen shot 2018-03-07 at 6 31 31 pm

broskoski:
Closes #1127 

collaborator

### DIFF
--- a/react/components/UI/Modal/Modal.js
+++ b/react/components/UI/Modal/Modal.js
@@ -15,7 +15,6 @@ const ModalBackdrop = styled.div`
   left: 0;
   background-color: rgba(255, 255, 255, 0.9);
   z-index: 6001;
-  transform: translate3d(0, 0, 1px);
 `;
 
 export default class Modal extends Component {
--- a/react/components/UI/Modal/Modal.js
+++ b/react/components/UI/Modal/Modal.js
@@ -15,6 +15,7 @@ const ModalBackdrop = styled.div`
   left: 0;
   background-color: rgba(255, 255, 255, 0.9);
   z-index: 6001;
+  transform: translate3d(0, 0, 1px);
 `;
 
 export default class Modal extends Component {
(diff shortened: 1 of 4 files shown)


react/components/UI/ModalDialog/index.js
```diff
@@ -8,7 +8,7 @@ const Dialog = styled.div`
   width: 90%;
   height: 90%;
   max-width: 40em;
-  max-height: 60em;
+  max-height: calc(100vh - 7em);
```
dzucconi: This is gonna look odd on super tall screens. If this is a workaround for Safari there's a better way (I don't know off hand)

react/components/UI/Modal/Modal.js
```diff
@@ -15,6 +15,7 @@ const ModalBackdrop = styled.div`
   left: 0;
   background-color: rgba(255, 255, 255, 0.9);
   z-index: 6001;
+  transform: translate3d(0, 0, 1px);
```
dzucconi: ?

dzucconi: If you're trying to force GPU then use `transform: translatez(0);` —this is almost certainly going to cause aliasing bugs.

broskoski: Moves it up one pixel on the z-axis

dzucconi: Also why

react/components/UI/ModalDialog/index.js
dzucconi: Fine for now though. But I'd leave a note about the underlying issue: Safari flexbox blah blah

react/components/UI/Modal/Modal.js
dzucconi: I know what it does but that doesn't explain anything.

broskoski: We do this on the header bar to fix position:fixed bugs on mobile safari

dzucconi: Did you actually experience an issue on Mobile Safari? Can you document why it needs to be 1px in a comment pointing to some external source? Can we move it into a media query that only executes on mobile devices.

PR #1171, opened 2018-03-28
Add find original link for text

broskoski:
For your consideration:

screen shot 2018-03-28 at 2 29 08 pm


broskoski:
cc @dshipper

PR #1183, opened 2018-04-06
Use dismisser instead of cookies lib for onboarding tips

broskoski:


components/tips/view.coffee
```diff
@@ -8,7 +8,10 @@ module.exports = class TipView extends Backbone.View
     'click': 'onClick'
     'click .js-close' : 'close'
 
-  initialize: ({ @user }) -> #
+  initialize: ({ @user }) ->
+    @dismisser = new Dismisser
+      key: @model.id
```
dzucconi: What's model id evaluate to?

broskoski: Something like so: https://github.com/aredotna/ervell/blob/master/apps/user/tips.coffee#L3

dzucconi: ah cool. got it

components/tips/view.coffee
```diff
@@ -52,6 +55,9 @@ module.exports = class TipView extends Backbone.View
         @remove()
 
   remove: ->
-    Cookies.set @model.id, true
+    @dismisser.dismiss()
```
dzucconi: So wait where is the code that prevents the tips from loading

broskoski: https://github.com/aredotna/ervell/blob/master/apps/user/tips.coffee#L3

broskoski: https://github.com/broskoski/ervell/blob/77cfbd74d8fa5b9523876942ef8f1a8ae20d3047/components/tips/helpers.coffee#L4

broskoski: Yeah, its a little weird. Tips get set on the route level (bc different tips for feed and for profile), so this adds all the tips that the route is trying to set, and rejecting the ones that already have cookies.

dzucconi: It is odd. If it works it works. It'd probably be clearer to pass the dismissers into the templates or something, then use `if tip[id].dismisser.dismissed()`.

If this works for now that's fine as is.

broskoski: Yeah that would be clearer, personally I think we should replace these entirely with a different approach.

dzucconi: Yeah we will. You can access local data like cookies through an Apollo datalink, setting up a schema for it and so forth. So we'll go that route when we switch these pages over.

PR #1191, opened 2018-04-09
More specific rules for disallowing indexing

broskoski:
- [x] no index on following and follower pages
- [x] no index on block pages when any of the connected channels are owned by a user with indexing turned off


apps/block/routes.coffee
```diff
@@ -19,11 +19,14 @@ _ = require 'underscore'
     res.locals.sd.BLOCK = block.toJSON()
     res.locals.sd.COMMENTS = comments.toJSON()
     res.locals.sd.CURRENT_ACTION = 'block'
+    canIndexConnections = _.all block.get('connections'), (c) -> c.can_index is true
```
broskoski: True is true

PR #1206, opened 2018-04-17
Make invest CTA show up above invite message

broskoski:


components/invest_cta/index.styl
```diff
@@ -4,19 +4,19 @@
...
-  padding 1em
+  padding 1.25em
 
   display flex
   flex-direction row
   justify-content center
   align-items center
   text-align center
-  z-index header_z
+  z-index header_z + 10
```
dzucconi: world war z-index

broskoski: Zombie CTA's coming over the wall

Issue #1198, opened 2018-04-12
new header metadata design + new edit channel modal


csherron:
CHANNELS

edit channel modal
https://zpl.io/blm1YN1

private channel
https://zpl.io/aBEj0NK

public channel expanded
https://zpl.io/agn9JN1

closed channel truncated
https://zpl.io/V4JPBvz

private channel (small window, stacked)
https://zpl.io/brrmZW3

PROFILE ETC

profile (somebody else)
https://zpl.io/anm1JKr

profile (your own)
https://zpl.io/boBYXDX

explore
https://zpl.io/bLwBAKM

DETAILS

expanding columns detail
https://zpl.io/2j8kJQp

metadata buttons detail
https://zpl.io/25yeBOo

csherron:
- [ ] for editing a profile bio, add form in user settings "About" (helper text "Write your bio here"):

![screen shot 2018-05-01 at 1 21 17 pm](https://user-images.githubusercontent.com/3780330/39484286-438261a4-4d43-11e8-9d03-a5042fdfbd93.png)

also add section headers 
- [ ] "Info"
- [ ] "About"
- [ ] "Password"

Issue #1154, opened 2018-03-23
private channels with public links available should have persistent link under channel info


csherron:
this needs to persist when a link has been made

<img width="250" alt="screen shot 2018-03-23 at 3 57 06 pm" src="https://user-images.githubusercontent.com/3780330/37850789-e52f2910-2eb2-11e8-88c1-4d760b4ea8b9.png">

dzucconi:
Can you explain the reasoning behind this? Seems odd to me

PR #1226, opened 2018-04-24
[WIP] Header / Metadata

dzucconi:
Closes https://github.com/aredotna/ervell/issues/1198

- [x] Wire up share button
- [x] Wire up follow button
- [x] Wire up mute button
- [x] Wire up edit channel button
- [x] Implement edit channel modal
- [x] Refine font-sizing, etc.
- [x] Wire up export options
- [x] Hover expand
- [x] Connect
- [x] Connect: Ticker-tape style hover
- [x] Connect: Opacity dip on select
- [x] Connect: debounce / persist results while loading
- [x] Minimize fetching on channel initial channel load (mock out skeleton response with GraphQL)
- [x] https://github.com/aredotna/ervell/issues/1245
- [x] Non-bold slash path separators
- [x] Check gradient support on Safari
- [x] https://github.com/aredotna/ervell/issues/1154
- [x] Invert gradient condition
- [x] Share link error handling
- [x] [Collapse top share link margin when there is no info](http://static.damonzucconi.com/_capture/g6pThoC3Dc.png)
- [x] Load new header on the various different filtered pages
- [x] Better handling of filter state on profiles
- [x] Update for can field change
- [x] Untangle the ridiculous rats nest of user routes
- [x] Re-initialize the legacy client
- [x] Sticky path
- [x] Mobile optimization
- [x] Right padding on metadata pockets
- [x] Adjust metadata pocket line vertical margin
- [x] Fix header search focus stacking with sticky path z-index
- [x] Header centering in Safari on profile pages that lack info panel
- [x] Put back tips
- [x] Followers/following in channel breadcrumb
- [x] Remove extra 's' in filter placeholder for blocks page
- [x] "Start private channel" link
- [x] Explore metadata
- [x] Put em dashes back
- [x] Some kind of empty message when you don't have any recent connections
- [x] "(+n)" in channel breadcrumb
- [x] Edit button should display for collaborative channels (but with different/limited options)?
- [x] Connected to shouldn't be shown if you are logged out and there are no connections (but always if you are logged in)

### API
- [x] Share links cannot be generated by collaborators
- [x] Return 0/0 follow counts for id is 21 (for some reason ...TBC)
- [x] Message channel mutation
- [ ] Non-premiums can mute


dzucconi:
- [ ] Wire up share button
- [ ] Wire up follow button
- [ ] Wire up mute button
- [ ] Wire up edit channel button
- [ ] Implement edit channel modal
- [ ] Refine font-sizing, etc.

dzucconi:
sharing

react/components/ChannelMetadata/components/ChannelMetadataExpandable/index.js
```diff
@@ -0,0 +1,58 @@
...
+      </Expandable>
+    )}
+  </Context.Consumer>
+);
+
+ChannelMetadataExpandable.propTypes = {
+  children: PropTypes.node.isRequired,
+};
+
+export default ChannelMetadataExpandable;
```
dzucconi: truncating-linked

New Context API is cool

dzucconi:
I think I've got enough of this working correctly for us to be in a position to not use the legacy connect in the react codebase.

connecting

lib/setup.coffee
```diff
@@ -128,7 +129,12 @@ module.exports = (app) ->
   switch NODE_ENV
     when 'development'
       reloadAndMount = createReloadable(app, require)
-      app.use reloadAndMount path.join(__dirname, '..', 'apps')
+
+      modules = glob.sync('./react/**/*.js').map (name) => path.resolve(name)
+
+      app.use reloadAndMount(path.join(__dirname, '..', 'apps'), {
+        watchModules: modules
```
damassi: Oh interesting -- I hadn't thought about this use-case when implementing this, pointing to internal code

dzucconi: Actually, I take that back. It doesn't appear to work.

damassi: In that case you can just `reloadAndMount(path.join(__dirname, 'react')`

damassi: I'll give this a look today, i think the assumed `node_modules` path under `watchModules` is what's breaking it

dzucconi: Appreciate it, sort of works—

An example:

I have a parent-component that gets rendered server-side. I make a change to a sub-component within that parent-component and HMR picks it up and injects it into the browser. But if I refresh, I do not see the changes. Only when I hit save on the parent-component, then I see the changes reflected on the server-side rendered instance.

dzucconi: Also in this instance, express-reloadable does log out a notice that the sub-component has changed

damassi: Can you point me to the exact parent / child? I've noticed this issue in Force when making changes to jade files rendered clientside (seems to be a bug in the webpack loader), but that only applies to that file type, and only sometimes

dzucconi: Sure, if you consider:

parent: https://github.com/dzucconi/ervell/blob/b4b880d7c2883d6b2c52a9d2219d33fa2edef23a/react/components/Channel/index.js

child: https://github.com/dzucconi/ervell/blob/b4b880d7c2883d6b2c52a9d2219d33fa2edef23a/react/components/ChannelMetadata/index.js

(or anything under within that)

The parent is called on the server here: https://github.com/dzucconi/ervell/blob/b4b880d7c2883d6b2c52a9d2219d33fa2edef23a/apps/channel/routes.coffee#L23

damassi: Can you try adding [`recursive: true`](https://github.com/artsy/express-reloadable/blob/master/index.js#L13) to the options?

dzucconi: lol: Once I add that if I make a change all of a sudden I get Apollo network errors on every reload. Removing `recursive: true` and everything is fine.

damassi: Damn... 🤔

dzucconi: For now at least just saving the parent file once is much faster than doing a full server restart/recompile!

damassi: Oh cool, its still possible to run Ervell locally, was able to get it booted -- gonna look at those two spots

damassi: Ok one last little thing and if that doesn't work not too sure for now (can also glob it instead of the extra `filter`, but you get the idea):

Re-add `recursive: true` and then 

```
    modules = glob.sync('./react/**/*.js')
      .filter (name) => !name.includes('react/apollo') # or name.includes('react/components')
      .map (name) => path.resolve(name)
```

dzucconi: No dice. Still getting the same "Network error: only absolute urls are supported"

damassi: Ah bummer, I'll keep an eye out for this in Force and if make sure to open a PR over here if I find a fix. Unfortunately `express-reloadable` is kinda a leap of faith since its messing with module caches directly without knowledge of what's going on in other libs

react/components/HomeMetadata/components/HomeBreadcrumb/index.js
```diff
@@ -0,0 +1,77 @@
...
+            '/explore/blocks': exploreOptions,
+            '/notifications': 'Notifications',
+          }[pathname]}
+        </StickyBreadcrumbPath.Crumb>
+      </StickyBreadcrumbPath>
+    );
+  }
+}
+
+export default WithCurrentRoute(HomeBreadcrumb);
```
dzucconi: We: really need a router. And: these need to be separate apps.

damassi: I would recommend looking at [found](https://github.com/4Catalyzer/found). Its very flexible

dzucconi: Weirdly, i can't find a comparison with react-router. Have you used both?

damassi: Yeah -- This is basically a more customizable react router 3, with much of the same API. React router 4 was a complete API redesign which made static route declarations 2nd class and thus optimizations like bundle splitting at the route level, etc, kinda hacky. For most common usecases tho it's basically RR3, just with all of its internals exposed should you want to adapt it to complex app env.

PR #1242, opened 2018-05-01
Clear notifications (but don't mark them as read) when dropdown is open

broskoski:
Fixes an annoying issue where notifications are not cleared when you click on an item (because the XHR is intercepted). This calls `clear` when the notifications are loaded, and "marks" them later.

### DIFF
--- a/collections/notifications.coffee
+++ b/collections/notifications.coffee
@@ -21,6 +21,9 @@ module.exports = class Notifications extends Feed
     @addOrMergeNew data
     mediator.trigger 'notification:added'
 
+  clear: ->
+    $.post "#{sd.API_URL}/notifications/clear"
+
   markRead: ->
     mediator.shared.current_user.resetNotificationCount()
     @each (group)->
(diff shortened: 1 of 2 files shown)


broskoski:
Closes #1240

dzucconi:
I totally didn't and don't understand this PR?

Also I'm realizing we can probably prefetch notifications when there is some unread then aggressively cache them. Which could give us the best of both worlds.

broskoski:
When notifications load, mark them as read on the server, but not the client. For instances when you click on a notification link, before this PR it would make an XHR request, which would then be interrupted when the new page loaded.

dzucconi:
Why not the client?

broskoski:
So that you can see which notifications are new....

broskoski:
It doesn’t work any differently now than it did before the PR. It just makes sure that notifications are marked as read when they load.

broskoski:
(On the server)

dzucconi:
The count doesn't go away when you mouseout now.

broskoski:
If that's the case it's not intentional (the code to clear it is still there): https://github.com/aredotna/ervell/pull/1242/files#diff-8e9ec0b3eef637b059fa94e19a2fc316R23

dzucconi:
Wherein there shall be no more discussion of Backbone View related code.

Issue #1227, opened 2018-04-24
Notifications should also dismiss on click


dzucconi:
## Steps to replicate
* Have an unread notification
* Hover the pulldown
* Click the notification

## Expected result
Marks notifications as read

## Actual result
It doesn't because the trigger is only on mouseout

PR #1248, opened 2018-05-07
Mark notifications as read first and then set the unread count

broskoski:
Closes https://github.com/aredotna/ervell/issues/1227 for good.


dzucconi:
Why though?

broskoski:
It looks like setting the state triggers a re-render, which interrupts that manually bound event in `initialize`

dzucconi:
🍰

PR #1262, opened 2018-05-21
Router fix for /welcome pages

broskoski:
When we send people "Welcome to Are.na" emails, we link to the Quick Guide (with UTM params). This loosens up the blacklist to allow them through, and ignores the params on the client-side router.

### DIFF
--- a/apps/onboarding/client/router.coffee
+++ b/apps/onboarding/client/router.coffee
@@ -11,6 +11,8 @@ module.exports = class OnboardingRouter extends Backbone.Router
     'welcome/:scene': 'scene'
 
   initialize: ->
+    @route /welcome((\?)(.)+)?$/, 'welcome'
+
     @user = new CurrentUser CURRENT_USER
     @state = new OnboardingState
 
@@ -19,6 +21,8 @@ module.exports = class OnboardingRouter extends Backbone.Router
 
   exec: ->
     @navigate "/welcome/#{@state.get 'scene'}", trigger: true
+
+  welcome: -> @scene 'welcome'
 
   scene: (scene) ->
     @state.set 'scene', (scene or 'welcome'), silent: true


apps/onboarding/client/router.coffee
```diff
@@ -11,6 +11,8 @@ module.exports = class OnboardingRouter extends Backbone.Router
     'welcome/:scene': 'scene'
 
   initialize: ->
+    @route /welcome((\?)(.)+)?$/, 'welcome'
```
dzucconi: ?

dzucconi:
We are very clearly doing something very wrong. Next time there is an issue with this can we just fix the dumb router?

apps/onboarding/client/router.coffee
broskoski: Catches any route with `?param=val` and sends it to the `welcome` route.

broskoski:
This thing (/welcome) won't exist for much longer, but yes.

apps/onboarding/client/router.coffee
dzucconi: I don't follow, why is this needed. Backbone router should just ignore the params and route to the correct path?

dzucconi: I mean realistically if it's not loading it was just cuz of the router conflict caused by the fucking dueling routers with the blacklist not matching

dzucconi:
Welcome isn't the thing. Global block router is the thing. But yes that also isn't long for this world.

apps/onboarding/client/router.coffee
broskoski: It does not ignore the params, they get set as a `:scene`

broskoski: https://github.com/jashkenas/backbone/issues/891

dzucconi: That is insane.

PR #1282, opened 2018-05-29
Fix following route

broskoski:
Feel free to suggest a nicer way to do this, but this is the basic idea.


broskoski:
This is where the client was being set up wrong: https://github.com/aredotna/ervell/blob/master/apps/user/client/index.coffee#L54

PR #1315, opened 2018-06-11
Sets up Storybook

broskoski:
Thinking this might be useful, sets up both Storybook and mocking Apollo responses for components.

Comments inline. Feel free to reject this if it doesn't line up with how we see developing components or doesn't belong in this repo.


react/.storybook/config.js
```diff
@@ -0,0 +1,34 @@
...
+
+const mocks = {
+  Query: () => {
+    return {
+      status: () => {
+        return "Up"
+      },
+    }
+  },
+  Channel: () => {
```
broskoski: We can obviously get much more interesting/organized about mocks, this is just to get the ball rolling. See: https://www.apollographql.com/docs/graphql-tools/generate-schema.html#example

react/components/UI/BreadcrumbPath/index.js
```diff
@@ -11,6 +11,7 @@ const Container = styled.div`
   margin: 0 0 2em 0;
   font-size: ${styles.Type.size.h3};
   line-height: ${styles.Type.lineHeight.compact};
+  font-family: ${styles.Type.font.sans};
```
broskoski: Interestingly, most of these React components rely on inherited styles from Ervell. This makes me wonder if we should try and detach that dependency and duplicate styles on the actual components.

react/stories/channel.js
```diff
@@ -0,0 +1,45 @@
+import React, { Component } from 'react';
+import { Query } from 'react-apollo';
+import gql from 'graphql-tag';
+import { storiesOf } from '@storybook/react';
+import PropTypes from 'prop-types';
+import channelMetadataFragment from 'react/components/ChannelMetadata/fragments/channelMetadata';
+import ChannelBreadcrumb from 'react/components/ChannelMetadata/components/ChannelBreadcrumb';
+import ChannelMetadataInfo from 'react/components/ChannelMetadata/components/ChannelMetadataInfo';
+
+class MockChannel extends Component {
```
broskoski: Again, probably a nicer way to do this, just offering an example.

broskoski:
screen shot 2018-06-11 at 8 25 57 am

react/components/UI/BreadcrumbPath/index.js
dzucconi: On base components like buttons/forms/etc I've been just using all: initial. https://github.com/aredotna/ervell/blob/f5dcd4811e8b05542e303042b6be8c74299e6b5e/react/components/UI/GenericButton/index.js#L12 to reset. I'm not terribly concerned about most other things as we'll just fix them as they crop up. But if you want isolation; that's one way to go.

broskoski: If we want to use storybooks, that will probably be how things naturally evolve. Not sure if that's good or bad.

react/stories/channel.js
dzucconi: What's with the naming here? Be consistent with classes and their filenames.

react/stories/channel.js
```diff
@@ -0,0 +1,45 @@
+import React, { Component } from 'react';
+import { Query } from 'react-apollo';
+import gql from 'graphql-tag';
+import { storiesOf } from '@storybook/react';
+import PropTypes from 'prop-types';
+import channelMetadataFragment from 'react/components/ChannelMetadata/fragments/channelMetadata';
+import ChannelBreadcrumb from 'react/components/ChannelMetadata/components/ChannelBreadcrumb';
+import ChannelMetadataInfo from 'react/components/ChannelMetadata/components/ChannelMetadataInfo';
```
dzucconi: Try to logically group dependencies by type.

react/stories/channel.js
broskoski: This is not the focus of the file, just a helper component to fake the Apollo fetches.

dzucconi: Then I'd separate it out. One class per-file. I thought there was an eslint rule for this?

react/components/UI/BreadcrumbPath/index.js
dzucconi: Well it's good; we have to kill the old Stylus code. I just mean it'll happen eventually storybooks or not, and if you're building a UI primitive then you'll need to reset using `all: initial` anyway to prevent insane specificity wars from the competing Stylus.

react/stories/channel.js
```diff
@@ -0,0 +1,45 @@
...
+            return (<div />);
+          }
+          return (<ChannelComponent channel={data.channel} />);
+        }}
+      </Query>
+    );
+  }
+}
+
+storiesOf('Channel', module)
```
dzucconi: (Capitalize the filename)

react/stories/channel.js
dzucconi: Though generally this is sorta weird. Why not use the actual Channel component?

broskoski: I don't want the entire `ChannelMetadata` component, just want a query to fetch a mocked channel.

dzucconi: Mergeit

react/.storybook/config.js
```diff
@@ -27,8 +27,8 @@ addDecorator(
 );
 
 function loadStories() {
-  require('../stories/index.js');
-  require('../stories/channel.js');
+  require('../stories/UI.js');
```
dzucconi: Consistent with extensions, puhlease

dzucconi: (We do not specify them for .js)

broskoski: What does that mean

dzucconi:
Oh It just occurred to me we aren't linting this directory which is why I'm overhere nitpicking. Just enable lint plz

PR #1316, opened 2018-06-12
[WIP] New AddBlock component

broskoski:
Getting some work in progress up here. Getting this ready for new onboarding flow.

### DIFF
--- a/react/components/AddBlock/index.js
+++ b/react/components/AddBlock/index.js
@@ -1,0 +1,123 @@
+import React, { Component } from 'react';
+import styled from 'styled-components';
+import Styles from 'react/styles';
+import { Textarea } from 'react/components/UI/GenericInput';
+import { isURL } from 'validator';
+
+const Outline = styled.div`
+  width: ${Styles.Constants.blockWidth};
+  height: ${Styles.Constants.blockWidth};
+  background-color: ${Styles.Colors.gray.semiLight};
+  border-radius: .25em;
+  padding: 5px;
+  box-sizing: border-box;
+`;
+
+const Interior = styled.div`
+  background: white;
+  display: flex;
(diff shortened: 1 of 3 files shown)


react/components/AddBlock/index.js
```diff
@@ -0,0 +1,123 @@
...
+import { isURL } from 'validator';
+
+const Outline = styled.div`
+  width: ${Styles.Constants.blockWidth};
+  height: ${Styles.Constants.blockWidth};
+  background-color: ${Styles.Colors.gray.semiLight};
+  border-radius: .25em;
+  padding: 5px;
+  box-sizing: border-box;
+`;
```
dzucconi: We got a mixin for this. The trick is that done this way it'll mess up alignment. The mixin just avoids this issue by mimicking the outline property.

https://github.com/aredotna/ervell/blob/64fc8b4005f5ad68cbc96d01fed5869fd5fbcd36/react/styles/Mixins.js#L5-L17

react/components/AddBlock/index.js
```diff
@@ -0,0 +1,123 @@
...
+
+const Interior = styled.div`
+  background: white;
+  display: flex;
+  height: calc(100% - 0.5em - 2px);
```
dzucconi: Why the calc?

react/components/AddBlock/index.js
broskoski: Ah got it.

react/components/AddBlock/index.js
broskoski: Bc of the above alignment thing.

dzucconi:
Looks good to me. I always assume calc is a code smell but it might also be fine so feel free to ignore.

Issue #1304, opened 2018-06-04
Transfer channel (UI)


broskoski:
![edit channel modals](https://user-images.githubusercontent.com/821469/40923675-6b8207c6-67e3-11e8-9d73-edbad13aa64d.jpg)

csherron:
also see: zpl.io/adMWJXp

PR #1320, opened 2018-06-13
Channel Transfer UI

dzucconi:
Closes https://github.com/aredotna/ervell/issues/1304

### DIFF
--- a/react/components/ManageChannel/components/TransferChannel/index.js
+++ b/react/components/ManageChannel/components/TransferChannel/index.js
@@ -34,7 +34,7 @@ export default class TransferChannel extends Component {
     if (channel.is_pending_transfer) {
       return (
         <div>
-          {!channel.transfer_request.is_user_to_member &&
+          {channel.visibility === 'private' && !channel.transfer_request.is_user_to_member &&
             <Status>
               Note: {channel.transfer_request.user_to.name} does not have access to this channel.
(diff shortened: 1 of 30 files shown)


react/components/ManageChannel/components/TransferChannel/index.js
```diff
@@ -0,0 +1,61 @@
...
+
+  render() {
+    const { channel } = this.props;
+    const { query, debouncedQuery } = this.state;
+
+    if (channel.is_pending_transfer) {
+      return (
+        <Status>
+          {channel.transfer_request.user_to.name} needs to confirm the transfer.
+          After they confirm you will no longer own this channel.
```
broskoski: Can we add a message here if the transferee does not have access to the channel (private and transferee is not a collaborator)?

> Note: {user} does not have access to this channel. Consider temporarily adding them as a collaborator so they can view the channel before they accept the transfer

dzucconi:
# TODO

- [ ] Add a message if the transferee does not have access to the channel (private and transferee is not a collaborator)?
> Note: {user} does not have access to this channel. Consider temporarily adding them as a collaborator so they can view the channel before they accept the transfer

dzucconi:
OK, this half is done; deploying to staging. We're just missing confirmation emails for the time being.

PR #1321, opened 2018-06-13
[WIP] Content discovery page

broskoski:
screen shot 2018-06-13 at 7 30 41 pm

TODO
- [ ] Client-side setup for channel and block actions
- [ ] Pagination
- [ ] Show action buttons on channel thumbs by default
- [ ] Randomize order? (I just want this)


apps/examples/index.coffee
```diff
@@ -0,0 +1,65 @@
...
+    
+      graphQL(send)
+  )
+  .then (results) ->
+    each results, (result, i) ->
+      examples[i].channels = result.channels
+   
+    res.locals.examples = examples
+    res.render 'index'
+  .catch(next)
```
dzucconi: Not saying you have to but you might want to consider not writing CoffeeScript where we don't have to; like routes for instance. You can use the Apollo middleware's client to make queries: https://github.com/aredotna/ervell/pull/1320/files#diff-9e5e66872800e1139ae9aab6f79b83b0R36

broskoski: Ah k, will do.

apps/examples/client/example_channel_view.coffee
```diff
@@ -0,0 +1,12 @@
...
+{ default: FollowButton } = require '../../../react/components/FollowButton'
+
+module.exports = class ExampleChannelView extends Backbone.View
+
+  initialize: ->
+    console.log('initializing', @$el.data('id'))
+    @initializeFollowButton()
+
+  initializeFollowButton: ->
+    mountWithApolloProvider FollowButton, { id: @$el.data('id'), type: 'CHANNEL' }, $('.js-example-channel-follow')
```
dzucconi: Avoid using views unless they are directly responding to a model. Otherwise just use jQuery and scope the els appropriately.

apps/examples/client/example_view.coffee
```diff
@@ -0,0 +1,26 @@
...
+ 
+template = -> require('../templates/example.jade') arguments...
+
+module.exports = class ExampleView extends Backbone.View
+
+  events: 
+    'click .js-load-more' : 'fetchAndRender'
+
+  initialize: ->
+    @initializeSubViews()
```
dzucconi: Generally avoid calling code that renders in initialize.

apps/examples/client/example_view.coffee
```diff
@@ -0,0 +1,26 @@
...
+        el: $(this)
+
+  fetchAndRender: (e) ->
+    e.preventDefault()
+    
+    id = @$el.data('id')
+    
+    $.get "/api/examples/#{id}", (example) =>
+      @$el.html template
+        example: example
```
dzucconi: This whole thing doesn't look like a view to me either actually. Would be much simpler to read as just normal JavaScript. You're sort of just using views as a way to scope selectors which makes this extremely confusing. Just scope it and that's that.

apps/examples/client/index.coffee
```diff
@@ -0,0 +1,32 @@
...
+    totalPages = $(this).attr('data-total-pages')
+
+    $.get "/api/examples?page=#{currentPage}", (examples) ->
+      html = examplesTemplate examples: examples
+      $('.Page--examples-list').append html
+
+      initializeViews()
+
+      if currentPage >= totalPages
+        $('.js-load-more-examples').css display: 'none'
```
dzucconi: `.hide()`

apps/examples/client/index.coffee
```diff
@@ -0,0 +1,32 @@
...
+      else
+        $('.js-load-more-examples').attr('data-current-page', currentPage)
```
dzucconi: Why is this storing data in the DOM?

apps/examples/components/channel_group/index.jade
```diff
@@ -0,0 +1,13 @@
...
+    .ChannelBlockGroup__title
+      h4.Type.Type--size-h4
+        | Are.na / #{channel.user.name} /&nbsp;
+        span(class='Color--channel-#{channel.visibility}')!= channel.title
+    .ChannelBlockGroup__follow.js-example-channel-follow(data-id=example.id)
+      //- rendered on the client
+  .ChannelBlockGroup__inner.js-channel
+    if channel.blocks
+      for block in channel.blocks
+        include ../../../../components/block_v2/templates/block
```
dzucconi: Try to use some whitespace. These jade views are always difficult to read.

apps/examples/examples.coffee
```diff
@@ -0,0 +1,160 @@
...
+    description: 'Speculations and theories on what comes next'
+    channel_ids: [113460, 15975, 50058]
+  },
+  {
+    id: 'information'
+    title: 'Visualizing Information'
+    description: 'Bottomless archives of maps, dataviz, and diagrams'
+    channel_ids: [70731, 13047, 8560]
+  }
+]
```
dzucconi: I'd consider storing this as JSON.

apps/examples/stylesheets/index.styl
```diff
@@ -0,0 +1,49 @@
...
+
+  &__channels
+
+    .ChannelBlockGroup
+      margin-bottom 4em
+
+    .ChannelBlockGroup__header 
+      margin 2em .5em 1em
+
+    .ChannelBlockGroup__title > h4.Type
```
dzucconi: Why not use the `&__` syntax with all of these? You also shouldn't be scoping blocks under other blocks like this.

apps/examples/templates/example.jade
```diff
@@ -0,0 +1,16 @@
...
+        h4.Type.Type--size-h4
+          strong #{example.title}:&nbsp;
+          = example.description
+    .Page--examples__channels
+      for channel in example.channels
+        include ../components/channel_group/index
+    if example.channel_ids.length > example.channels.length
+      .Page--examples__button.Utility--align-center
+        a.Button.Button--size-lg.Button--border-heavy.js-load-more( href='#' )
+          | See more #{example.id} channels
```
dzucconi: I don't really see how this is a modifier on the `.Page` class. Also variants should always be using the thing it's modifying: `.Page.Page--modifier`

components/block_v2/templates/block_no_kind.jade
```diff
@@ -0,0 +1,62 @@
...
+      .Block__caption__title
+        = block.title
+        - if(block.__typename === 'Attachment' && ['mp3', 'pdf'].indexOf(block.file_extension))
+          img(data-src="#{sd.IMAGE_PATH}iconic/type-#{block.file_extension}.svg" class="iconic iconic-sm")
+        - if(block.__typename === 'Embed')
+          img(data-src="#{sd.IMAGE_PATH}iconic/type-embed.svg" class="iconic iconic-sm")
+
+      .Block__caption__meta
+        .Block__caption__meta__author Added by #{block.user.name}
+        .Block__caption__meta__timestamp Updated #{block.updated_at}
```
dzucconi: Let us hope this page doesn't require significant maintenance! We should start a hitcounter for competing block implementations.

broskoski:
I'm gonna merge this so it can get up on staging for Chris and Leo to review. The link isn't live anywhere, so should be fine.

Issue #1327, opened 2018-06-18
Analytics for channel transfer


broskoski:
- [x] Event for channel transfer initiation
- [x] Event for channel transfer acceptance
- [x] Event for channel transfer rejection

PR #1332, opened 2018-06-22
Transfer analytics

dzucconi:
Closes https://github.com/aredotna/ervell/issues/1327

I got a lil in the weeds here looking trying to handle the routing. I need to tie the mutations to the pages directly then use the router on the server as well. But I'm leaving this alone for now.

Going to revisit this typography stuff with styled-system in a bit; same idea really though.


apps/actions/client/index.js
```diff
@@ -0,0 +1,39 @@
...
+const { data: { APOLLO, STATUS_CODE } } = sharify;
+
+module.exports = () => {
+  const mountPoint = document.getElementById('apolloMount');
+
+  const Component = {
+    UNAUTHORIZED: TransferChannelAccessDeniedPage,
+    NOT_FOUND: TransferChannelNotFoundPage,
+    OK: Routes,
+  }[STATUS_CODE];
```
dzucconi: This is just a quick hacky way of dealing with the errors for now.

apps/actions/templates/index.jade
```diff
@@ -0,0 +1,13 @@
...
+  != apollo.styles
+
+block body
+  #apolloMount
+    != apollo.html
+
+block scripts
+  script( src= asset('/assets/actions.js') )
+  script.
+    window.__APOLLO_STATE__ = !{JSON.stringify(apollo.state).replace(/</g, '\\u003c')};
```
dzucconi: I need to write a JSX layout.

dzucconi: Though I suppose we'd need a new search/user bar to do that.

broskoski: Yeah true.

broskoski:
Looks great to me 💯

PR #1340, opened 2018-06-26
Add blog section to homepage

broskoski:
screen shot 2018-06-26 at 1 11 53 pm


lib/truncate.coffee
```diff
@@ -0,0 +1,8 @@
+{ take } = require 'underscore'
+
+module.exports = (text, length = 50) ->
+  return unless text
+  tokens = text.split ' '
+  truncated = take(tokens, length).join(' ')
+  truncated = truncated + "..." if tokens.length > length
+  truncated
```
dzucconi: `{ truncate } = require 'underscore.string'`

broskoski: Our truncate ensures that words don't get cut off, underscore.string does it by the character.

dzucconi: works for me

dzucconi: but i still dunno why you writing coffeescript these days

broskoski: Cause I'm a maniac

broskoski: This was actually already written, I just put it in its own file.

dzucconi: tru

PR #1342, opened 2018-06-26
Big annoying UI refactor

dzucconi:
Still working thru some things — much of this isn't using some of the system functions yet; rather just calling out to the injected theme.

<del>Some things bug me; like setting defaults in mixins is basically impossible.</del> I'm sort of on the fence about just writing my own style functions but it's sort of nitpicky and is relatively easy to change.

Refactoring this stuff is just bulk find/replaces. Going to add a bunch more stuff to the storybook real quick.


react/styles/theme.js
```diff
@@ -0,0 +1,42 @@
...
+  <ThemeProvider theme={THEME} {...rest}>
+    {children}
+  </ThemeProvider>
+);
+
+Themed.propTypes = {
+  children: PropTypes.node.isRequired,
+};
+
+export default THEME;
```
broskoski: Noice

react/components/UI/Text/index.js
```diff
@@ -0,0 +1,27 @@
+import styled, { css } from 'styled-components';
+import { display, space, textColor, fontFamily, fontSize, lineHeight, textAlign } from 'styled-system';
```
damassi: very happy about this

alloy:
Such style ✨

react/components/UI/Text/index.js
dzucconi: Good lookin out

react/components/UI/Text/index.js
```diff
@@ -0,0 +1,21 @@
...
+import { preset } from 'react/styles/functions';
+import { antialiased, neutralMarginsY } from 'react/styles/mixins';
+
+export const mixin = css`
+  ${display}
+  ${space}
+  ${preset(fontFamily, { font: 'sans' })}
+  ${preset(fontSize, { f: 5 })}
+  ${preset(lineHeight, { lineHeight: 1 })}
+  ${preset(textColor, { color: 'gray.base' })}
```
dzucconi: This is the only real gotcha I've run into. When you specify preset values; you'll want to use the shorthand aliases.

I explain here: https://github.com/jxnblk/styled-system/pull/37#issuecomment-400699999 —there might be a way around this by inspecting the function propTypes inside of the preset function I wrote. Or there might be some other obvious way to deal with this issue. But for now it's a minor inconvenience.

https://github.com/jxnblk/styled-system/blob/master/src/styles.js serves as a reference for aliases.

dzucconi: Another option would be to expose our own wrapped functions that accept presets.

dzucconi:
Am happy—merging.

dzucconi:
We need to set up style-guide deploys as well. I suspect this can be very useful for @csherron

PR #1347, opened 2018-06-29
Fix RSS templates

dzucconi:


apps/rss/routes.coffee
```diff
@@ -28,22 +28,21 @@ Posts = require "../../collections/posts.coffee"
   .done()
 
 @userRSS = (req, res, next) ->
-  author = new User
-    channel_slug: req.params.username
```
dzucconi: 🚨Copypasta Alert 🚨

dzucconi:
I don't think these set the right headers either. FWIW.

Issue #866, opened 2017-08-22
Validate email addresses on invite form


broskoski:
# Steps to reproduce

Enter an email address with a  "user name <name@email.com>"

# Expected result

Text input should return an error message

# Actual result

Invite gets sent

bradley:
Done.

PR #1365, opened 2018-07-11
Issue 866: Validate email addresses on invite form

bradley:
# Overview

References [#866](https://github.com/aredotna/ervell/issues/866).

- [x] Add catch-all error handling in case server responds with 500 (can happen if the `type='email'` on the frontend form ever fails to catch errant values).
- [x] Better timeout handling for UI updates to prevent prior requests from making UI updates on top of subsequent requests.
- [x] Not technically part of this PR, but I added Disposable Email checking in the sibling [API Pull Request](https://github.com/aredotna/api/pull/656) for this PR.

Notes:

This PR is marked WIP because of the following: 

Currently the existing code appears to stop the behaviour described in  [#866](https://github.com/aredotna/ervell/issues/866), at least in major browsers, because of the `type='email'` on the frontend form's input element, which makes me think [#866](https://github.com/aredotna/ervell/issues/866) needs more explanation. 

That said, I do think that relying on `type='email'` can be unreliable and is better off to move away from. To that end, I added some questions on the sibling [API Pull Request](https://github.com/aredotna/api/pull/656) for this PR.

If we want to do more validation server side, the changes to the form in this PR should accommodate them.

# Screenshots

### Chrome, Default Email Input Behaviour
image

### Firefox, Default Email Input Behaviour
image

### Disposable Email Handling
image

### Catch-All (500 etc) Error Handling
image

### DIFF
--- a/apps/tools/components/send_invitation/index.coffee
+++ b/apps/tools/components/send_invitation/index.coffee
@@ -8,8 +8,12 @@ module.exports = ($el) ->
   $submit = $el.find '.js-submit'
   $errors = $el.find '.js-errors'
 
+  submissionTimeout = null;
+
   $form.on 'submit', (e) ->
     e.preventDefault()
+
+    submissionTimeout && clearTimeout(submissionTimeout);
 
     serializer = new Serializer $form
 
@@ -31,7 +35,7 @@ module.exports = ($el) ->
         .prop 'disabled', false
         .text 'Sent!'
 
-      setTimeout (-> $submit.text label), 2500
+      submissionTimeout = setTimeout (-> $submit.text label), 2500
 
       track.submit 'Invitation sent from user'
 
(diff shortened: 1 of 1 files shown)


apps/tools/components/send_invitation/index.coffee
```diff
@@ -8,12 +8,42 @@ module.exports = ($el) ->
...
+      $errors.empty()
+    , 5000
+
+    track.error 'Invitation not sent, try again.'
+
   $form.on 'submit', (e) ->
     e.preventDefault()
 
-    serializer = new Serializer $form
+    submissionTimeout && clearTimeout(submissionTimeout);
```
dzucconi: Since this is a .coffee file we might as well be idiomatic; (I realize there's no linter here) (`&& => and`, no semi-colons).

Also `clearTimeout` will return fine on null/undefined, so there's no need to do the existence check.

apps/tools/components/send_invitation/index.coffee
```diff
@@ -24,32 +54,10 @@ module.exports = ($el) ->
...
-      setTimeout ->
-        $submit.text label
-        $errors.empty()
-      , 5000
-
-      track.error 'Invitation not sent, try again.'
+      onFailure("#{message}<br/>#{description}")
+
+    .catch ->
+      onFailure("Something went wrong, please contact <a href='mailto:info@are.na'>info@are.na</a> if the problem persists.")
```
dzucconi: Ah yeah, the catch-all is a good call. We've definitely encountered that in the wild.

bradley:
Note for the future: Might be worth adding some kind of catch for `429` responses in the `components/send_invitation/index.coffee` file to tell the user to slow down when they are mashing send repeatedly. Right now it will just fall back to the general "Something went wrong" message.

broskoski:
💯

PR #1367, opened 2018-07-12
Adds ID to channel can query

dzucconi:


dzucconi:
Re: https://github.com/apollographql/apollo-client/issues/2510

PR #1368, opened 2018-07-12
Receive newsletter

broskoski:
All this work for a checkbox:

screen shot 2018-07-12 at 11 37 23 am

Also a new setting here:
screen shot 2018-07-12 at 11 38 00 am


dzucconi:
+++++

PR #1371, opened 2018-07-12
Update /examples styling

broskoski:
examples-page

### DIFF
--- a/apps/examples/client/index.coffee
+++ b/apps/examples/client/index.coffee
@@ -9,7 +9,6 @@ module.exports = ->
   # 
   # Load more channels (single example category)
   # 
-
   $(document).on 'click', '.js-load-more', (e) ->
     e.preventDefault()    
     id = $(this).data('id')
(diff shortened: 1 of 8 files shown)


apps/examples/templates/index.jade
```diff
@@ -9,23 +9,14 @@ block scripts
...
     include examples
-  
-  .Page.Page--examples
-    .Page--examples__button-load-more.Utility--align-center
-      a.Button.Button--size-lg.Button--border-heavy.js-load-more-examples( href='#' data-current-page=1 data-total-pages=totalPages )
-        | See more examples
+
+    //- .Utility--align-center
+    //-   a.Button.Button--size-lg.Button--border-heavy.js-load-more-examples( href='#' data-current-page=1 data-total-pages=totalPages )
+    //-     | See more examples
```
dzucconi: Delete?

apps/examples/client/index.coffee
```diff
@@ -18,10 +17,20 @@ module.exports = ->
...
+
   # 
-  # Follow buttons
+  # Handle example channel click
   # 
+  $(document).on 'click', '.js-channel', (e) ->
+    e.preventDefault()
+    e.stopImmediatePropagation()
+    url = $(this).data('href')
+    window.open url
```
dzucconi: Why?

apps/examples/templates/index.jade
broskoski: Yes

apps/examples/client/index.coffee
broskoski: Don't open the lightbox when clicking on a block, just go to the channel

dzucconi: hOW vErY FasciNaTIng

PR #1372, opened 2018-07-12
Ensures the ID matches

dzucconi:
Noticed this was causing a rehydration error locally, but not in production.


dzucconi:
Sort of wondering if we should switch to doing something similar to what Artsy did/does, which is replace the ID field with the slug and return the actual ID under `_id`

Issue #1300, opened 2018-06-01
Search results page header/metadata update


dzucconi:
Needs to be updated to match.

broskoski:
Some more notes here: 

This is mostly just a styling update but a chance to pull in some of the new React components we've been building for new UI.

---

Here is how the current search header looks:
<img width="1015" alt="screen shot 2018-07-11 at 10 06 38 pm" src="https://user-images.githubusercontent.com/821469/42608639-b6bff65e-8556-11e8-85bc-11274939e4c7.png">

vs the Explore page:

<img width="991" alt="screen shot 2018-07-11 at 10 07 06 pm" src="https://user-images.githubusercontent.com/821469/42608652-cc72c90e-8556-11e8-82de-dbbeb4e4cf38.png">

---

So all that needs to happen here is replacing the jade template there (on the path and the 'view' column) to use the new components, found in: https://github.com/aredotna/ervell/tree/master/react/components/HomeMetadata

PR #1374, opened 2018-07-13
Issue 1300: Header/Metadata Update for Search

bradley:
# Overview

This PR closes issue [#1300](https://github.com/aredotna/ervell/issues/1300), updating the UI for search results and converting the code to utilize React.

### Notes
I noticed that a few places were defining their own styled components for the header metadata links (e.g.; the profile and explore pages) and went ahead and made a reusable UI component for this and updated the existing components.

# Screenshots

### Old
image

### New
image


apps/search/index.js
```diff
@@ -0,0 +1,78 @@
...
+const app = express();
+
+app.set('views', `${__dirname}/templates`);
+app.set('view engine', 'jade');
+
+const renderSearch = (req, res, next) => {
+  const blocks = new SearchBlocks();
+  const search = res.locals.sd.SEARCH;
+
+  if (req.params.block_id) {
```
broskoski: FYI, we don't support these urls anymore so this is fine to remove.

apps/search/index.js
```diff
@@ -0,0 +1,78 @@
...
+  setSearchMiddleware,
+  apolloMiddleware,
+  setHeaderMiddleware
+];
+
+app.get('/search/:query', ...middlewareStack, renderSearch);
+app.get('/search/:query/channels', ...middlewareStack, renderSearchChannels);
+app.get('/search/:query/blocks', ...middlewareStack, renderSearchBlocks);
+app.get('/search/:query/users', ...middlewareStack, renderSearchUsers);
+app.get('/search/:query/block/:block_id', ...middlewareStack, renderSearch);
```
broskoski: Ditto here

apps/search/index.js
bradley: 👍

apps/search/index.js
bradley: 👍

bradley:
@dzucconi @broskoski I think it may be worth considering adding some aliases into the webpack config. 

```
config.resolve.alias = {
  Components: path.join(__dirname, "react", "components"),
  UI: path.join(__dirname, react", "components", "ui")
};
```

This would allow `import` statements to be shorter in all of the react files. For example with the above aliases you go from this:

```
import Grid from 'react/components/UI/Grid';
import HomeBreadcrumb from 'react/components/HomeMetadata/components/HomeBreadcrumb';
```

to

```
import Grid from 'UI/Grid';
import HomeBreadcrumb from 'Components/HomeMetadata/components/HomeBreadcrumb';
```

It's not a huge pain without this, but makes reading paths easier and as well as removing the need for doing a lot of `../../some_concept/some_component` type of `imports`. It's a little weird with the naming convent of having most of the react components have their own internal `components` directory though.

broskoski:
Your recommendation sounds good to me, but I think @dzucconi should have the final word. Otherwise, well done! 👏 👏

apps/search/index.js
```diff
@@ -0,0 +1,71 @@
+import express from 'express';
+import SearchBlocks from '../../collections/search_blocks';
+import { data as sd } from 'sharify';
```
dzucconi: I don't think this is being used? You might want to turn on inline linting in your editor if possible. ESLint isn't running on CI yet.

apps/search/middleware/setHeader.js
```diff
@@ -0,0 +1,15 @@
...
+
+  return req.apollo.render(SearchComponent, {
+    search: SEARCH
+  })
+    .then((searchComponent) => {
+      res.locals.searchComponent = searchComponent;
+      next();
+    })
+    .catch(next);
+  next();
```
dzucconi: This next is unreachable and can be removed.

react/components/SearchMetadata/index.js
```diff
@@ -0,0 +1,36 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import { Route } from 'react-router-dom';
+import WithStaticRouter from 'react/hocs/WithStaticRouter';
+import Grid from 'react/components/UI/Grid';
+import HeaderMetadataContainer from 'react/components/UI/HeaderMetadata/HeaderMetadataContainer';
+
+import SearchMetadataBreadcrumb from './components/SearchMetadataBreadcrumb';
+import SearchMetadataModeView from './components/SearchMetadataModeView';
```
dzucconi: I'm cool with adding aliases for imports. (I like to copy paths of files out of my editor with shortcuts). Aliases or absolute paths are preferred though so that we can just copy imports indiscriminately between files.

apps/search/index.js
```diff
@@ -0,0 +1,71 @@
...
+import { extend } from 'underscore';
+import apolloMiddleware from 'react/apollo/middleware';
+import setModeMiddleware from './middleware/setMode';
+import setSearchMiddleware from './middleware/setSearch';
+import setHeaderMiddleware from './middleware/setHeader';
```
dzucconi: Relative paths here

dzucconi:
Looks great—few minor things I'd address then merge when green.

In the future I'd also consider splitting that WIP commit up and flattening it into some of the other commits it's related to, which you could do with an interactive rebase.

bradley:
👍 @dzucconi, sorry for the WIP commits.

PR #1380, opened 2018-07-19
Deploy search header and examples test

broskoski:


broskoski:
cc @bradley your header is getting deployed!

PR #1386, opened 2018-07-26
Fix mobile logged out menu

broskoski:


broskoski:
No need for scrolling down about page anymore, just go directly to the links.

Issue #1383, opened 2018-07-23
Onboarding flow


broskoski:
Replace `/welcome` route with this flow. The feed / find friends flow should be done separately (@broskoski will do).

---
![make a channel](https://user-images.githubusercontent.com/821469/43088362-f725604a-8e6f-11e8-8de2-a4589afe926e.jpg)

PR #1389, opened 2018-08-01
Issue 1383: New Onboarding Flow

bradley:
# Overview

This PR addresses [#1383](https://github.com/aredotna/ervell/issues/1383).

## Development Comments
- The animated transitions between steps are easy to control based on state using the `TransitionGroup`/`AnimatedPage` pattern you'll see in this PR. It meets the spec (i.e.; "Dissolve In/Out Animation on Everything") but we can tweak the animation however ya'll want. Right now it just uses state in the entrypoint, so "back" is not supported (e.g.; it will just go back to whatever page you were at prior to the entire `/welcome` history state), but if we'd like we can also easily do the same transitioning with `ReactRouter` given that we want URLs for each page. I didnt think that was what was wanted so went with the simpler state-based transitions.
- There is a discrepancy between the mockup for the Channel Name input and the implementation. Between initial development and now the shared UI component for Inputs (previously `GenericInput`) changed to remove border radii and other minor stylistic differences such as a somewhat darkened border. I brought this up with Charles and he felt it was fine but advised me to mention it to Chris in Basecamp. I have done so and if changes need to be made I can do so.

# Screenshots

#### Mockup
image

#### Implementation
image

#### Mockup
image

#### Implementation
image

#### Mockup
image

#### Implementation
image

#### Mockup
image

#### Implementation
image


react/components/Onboarding/components/AnimatedPage/index.js
```diff
@@ -0,0 +1,72 @@
...
+
+  animateOut(callback) {
+    this.setState({ show: false }, () => {
+      setTimeout(callback, ANIMATION_PERIOD)
+    });
+  }
+
+  render() {
+    return (
+      <VelocityComponent
```
dzucconi: I think Velocity is great; but I'm reluctant to introduce a dependency for something that could be simply handled with some CSS. We could handle the period with props inside of css as well thanks to styled-component allowing you to access them.

What do you think?

react/components/Onboarding/components/UI/CTAButton/index.js
```diff
@@ -0,0 +1,11 @@
+import theme from 'react/styles/theme';
+import GenericButton from 'react/components/UI/GenericButton';
+import styled from 'styled-components';
+
+const CTAButton = styled(GenericButton).attrs({
+  f: 6,
+})`
+  margin-top: ${theme.space[6]}
```
dzucconi: This can be `mt: 6` in the attrs declaration above.

Also generally you don't want to be accessing the theme directly but rather through the props `${x => x.theme.space[6]}` as this allows the theme to be swapped out dynamically (something we're not doing yet but I can see us doing for things like night-mode, for instance).

react/components/Onboarding/components/Welcome/index.js
```diff
@@ -0,0 +1,31 @@
...
+const Welcome = props => {
+  const { goForward } = props;
+
+  return (
+    <div>
+      <Text f={9} my={4}>
+        Welcome to Are.na!
+      </Text>
+      <Text f={9} my={4}>
+        Let's make your first channel.
```
dzucconi: Simple thing but use the unicode rsquo: `’` instead of the feet glyph.

dzucconi: @danpia @csherron When we prepare copy can we make sure we have the correct chars (I tend to have to remember to replace these things.)

danpia: Good point, should also be the case for existing copy, about page, etc. ccing leo
For curly single quote: `Alt + Shift + ]`

dzucconi: (I can never remember it because the html entities are burned into my memory and always copy paste from this network of sites http://www.rsquo.net/)

react/components/Onboarding/components/AnimatedPage/index.js
bradley: Sure Ill look into it tonight.

react/components/Onboarding/components/UI/CTAButton/index.js
bradley: > This can be mt: 6 in the attrs declaration above.

I looked at `GenericButton` and didnt think it implemented `space` from `styled-system`, which I think is required for `mt`. Ill try this though.

react/components/Onboarding/components/Welcome/index.js
bradley: 👍

react/components/Onboarding/components/UI/CTAButton/index.js
dzucconi: Yeah, a bunch of stuff got migrated in a bulk fashion and hasn't been updated yet; working on some of those now.

bradley: Yeah `mt: 6` in the `attrs` declaration doesnt seem to do anything, I believe because `space` is not yet part of `GenericButton`. `margin-top: ${x => x.theme.space[6]}` works though.

dzucconi: 👍

react/components/Onboarding/components/UI/CTAButton/index.js
```diff
@@ -0,0 +1,13 @@
+import { space } from 'styled-system';
+import { preset } from 'react/styles/functions';
+import styled from 'styled-components';
+import GenericButton from 'react/components/UI/GenericButton';
+
+const CTAButton = styled(GenericButton).attrs({
+  f: 6
+})`
+  ${space}
+  ${preset(space, { mt: 6 })}
```
bradley: Scratch that @dzucconi, this does it.

react/components/Onboarding/components/Channels/components/CreateChannel/index.js
```diff
@@ -114,52 +141,68 @@ class CreateChannel extends React.Component {
...
       </CreateChannelWrapper>
     );
   }
 };
 
-CreateChannel.propTypes = {};
+CreateChannel.propTypes = {
+  createChannel: PropTypes.func.isRequired,
+  data: PropTypes.shape({
+    loading: PropTypes.bool.isRequired,
```
dzucconi: To easily proptype out the data shape you can use fragments that are used in your queries, like so: https://github.com/aredotna/ervell/blob/fae48e5f288d77251acd688531f700d150a3d5ba/react/components/ProfileMetadata/index.js#L19

react/components/Block/index.js
```diff
@@ -0,0 +1,61 @@
...
+  return (
+    <BlockWrapper>
+      <BlockPadding />
+      {blockContent()}
+    </BlockWrapper>
+  );
+};
+
+Block.propTypes = {
+  blockData: PropTypes.shape({}).isRequired,
```
dzucconi: This should be validated with an actual data shape using the the proptype of the fragment used to fulfill it. 

a la
https://github.com/aredotna/ervell/blob/497d8f3bfbb1e6895dfa5745333f9359b1b0a9e0/react/components/ChannelMetadata/components/ChannelMetadataActions/index.js#L22

dzucconi: Oh, I see this is the like top-level block—I suppose it could be something like a oneOf(shape) sort of thing.

react/components/Onboarding/components/Channels/components/IntroduceChannel/index.js
```diff
@@ -0,0 +1,99 @@
...
+})`
+  text-align: center;
+`;
+
+
+class IntroduceChannel extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.imageSrc = '/images/onboarding/addblock.gif';
```
dzucconi: This seems odd to me? Assets should be fingerprinted in production so this may not work there. You could do a require (the webpack config for that might need to be updated, not certain).

react/components/Onboarding/components/UI/CTAButton/index.js
```diff
@@ -0,0 +1,13 @@
+import { space } from 'styled-system';
+import { preset } from 'react/styles/functions';
+import styled from 'styled-components';
+import GenericButton from 'react/components/UI/GenericButton';
+
+const CTAButton = styled(GenericButton).attrs({
+  f: 5,
+})`
+  ${space}
+  ${preset(space, { mt: 6 })}
```
dzucconi: You don't need to include space, nor set the preset—GenericButton implements space so you could just set mt in the attrs.

react/components/Onboarding/components/UI/ChannelTitleInput/index.js
```diff
@@ -0,0 +1,26 @@
...
+})`
+  max-width: 100%;
+
+  input {
+    // Note: Autosize input has box-sizing "content-box". Here we subtract
+    // padding in input and border width to account for this when calculating max-width.
+    max-width: calc(100% - ${inputHorizontalPadding} - ${inputHorizontalPadding} - 2px) !important;
+    ${mixin}
+    ${space}
+    ${preset(space, { my: 5 })}
```
dzucconi: No need to re-include space with a preset. And you only need to include space if you want it to be configurable in multiple situations, otherwise pull values out of the props or just use em.

dzucconi:
Looking good—glad the rebase was figured out. Let me know if you want to push this to staging at some point soon.

react/components/Block/index.js
bradley: Yeah this is more "any props you want to send to the thing being plugged into the block (e.g.; channel, video, image, URL, etc). It's kind of just a catch-all object. Not much to validate on it and should be done on the block type itself.

react/components/Onboarding/components/Channels/components/IntroduceChannel/index.js
bradley: Yeah. Looks like all of ya'lls images are cloudfront URLs (like [this](https://github.com/aredotna/ervell/blob/master/apps/about/templates/faqs.jade#L45)). I wasnt really sure if you wanted to host the image somewhere and give me the URL or what. I can look at webpack config to require.

react/components/Onboarding/components/UI/CTAButton/index.js
bradley: 👍

react/components/Onboarding/components/UI/ChannelTitleInput/index.js
bradley: 👍

bradley:
Yeah the rebase issue was just a [GitHub bug](https://stackoverflow.com/questions/16306012/github-pull-request-showing-commits-that-are-already-in-target-branch) causing the PR to show commits already in master. Let's push to staging once I get a chance to address the comments above.

dzucconi:
Can you remove package-lock.json from this and install using yarn instead of npm? I can't get this onto a server with conflicting lockfiles

dzucconi:
Alrighty—this is on https://staging.are.na/welcome now @csherron @broskoski

broskoski:
@Bradley some minor copy updates:

Page 1 (no change)
Welcome to Are.na! 
Let's make your first channel.

Page 2
Channels are collections of content built around an idea.
You can add images, links, PDFs, videos, text, and more.

Page 3 
First, type a channel name 
(for example: "bookmarks" or "inspiration")

Final Modal
Congrats!
Get started by adding files, links, or text to your channel.

bradley:
Added the copy changes we discussed and deleted `package-lock.json`

bradley:
I think this is ready for release unless you need me to do anything @dzucconi

dzucconi:
Looks good to me, waiting on @broskoski @csherron

dzucconi:
I think @broskoski wants to get some empty feed state in first.

broskoski:
@Bradley Looks great to me as well, and yes let's deploy when https://github.com/aredotna/ervell/pull/1418 is all wrapped up as well. Well done!

broskoski:
@Bradley just pulled this down for a final check, it looks like the webpack config is broken for the image. We can also upload the image to s3/cloudfront and skip the webpack config-ing for now if that's easier. Let me know what you prefer.

react/components/Onboarding/components/Channels/components/IntroduceChannel/index.js
```diff
@@ -0,0 +1,95 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+import Text from 'react/components/UI/Text';
+import addBlockImagePath from 'Images/onboarding/addblock.gif';
```
broskoski: Punting on webpack config would mean changing this to this:

```
const addBlockImagePath = 'https://d2w9rnfcy7mm78.cloudfront.net/2582718/original_fe96a805b0771ca9e44a235c7f9ea3cd.gif';
```

broskoski:
Gonna merge and hardcode this for now. Thanks @Bradley!!

PR #1390, opened 2018-08-02
[WIP] Auth

dzucconi:
- [x] Fix remaining UI fallout
- [x] Fix broken follows (some nonsense initialization)
- [x] Adapt signup form for accepting an invitation (this needs a corresponding query and mutation on the API as well)
- [x] Perform reset
- [x] Error handler needs to account for when there is no reasonable response (500s) (currently failing on prop attribute of null (extensions))
- [x] Fix refresh endpoint
- [x] Sigh fix confirmation app
- [x] Analytics


dzucconi:
reset

dzucconi:
login

PR #1391, opened 2018-08-03
Add Jest for tests

damassi:
Noticed in https://github.com/aredotna/ervell/pull/1390 that some tests were starting to get written so wanted to suggest Jest before things got too far along. This lib is really futuristic as far as test runners go and is basically zero config. It looks for files with a `file.test.js` extension (or files placed in a `__tests__` directory) so you can very easily co-locate tests alongside their sources which tends to dramatically increase test-coverage over time since everything is side-by-side. 

jest

(Also, not sure if you've used Enzyme before but it's great for writing [component tests](https://github.com/artsy/reaction/blob/master/src/Components/Publishing/Display/__tests__/DisplayPanel.test.tsx) and works nicely with Jest.)

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
     "predeploy": "sh scripts/predeploy.sh",
     "mocha": "sh scripts/mocha.sh",
     "test": "sh scripts/test.sh",
+    "jest": "jest",
     "start-dev": "foreman start -f Procfile.dev",
     "deploy-staging": "APP_NAME=ervell-staging S3_BUCKET=ervell-staging yarn deploy"
   },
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -141,5 +143,8 @@
   },
   "optionalDependencies": {
     "fsevents": "*"
+  },
+  "jest": {
+    "testURL": "http://localhost"
```
damassi: Can remove this once this [patch](https://github.com/facebook/jest/pull/6792) is published. JSDom (used underneath Jest) made a change upstream and the internet is losing its mind.

dzucconi:
Cool I'm with it. Will just add swap out Mocha in my branch and co-locate my spec for that other fn instead of dealing with this merge. 👍

package.json
dzucconi: thx much

PR #1398, opened 2018-08-09
Premium message split test

broskoski:
screen shot 2018-08-08 at 4 55 39 pm
screen shot 2018-08-08 at 4 55 00 pm

p.s. Not seeing any of the premium events in GA, however I do see them firing locally. Thinking that maybe the page was reloading before they can fire, but it is mysterious.


dzucconi:
Yeah, it's not going to be accurate unless we do the whole blocking url thing.

broskoski:
@dzucconi what about a short `setTimeout` promise, or something along those lines? Good idea or bad? It would obviously be nice to be able to see which test gets more premium.

dzucconi:
Yeah that's basically what Mixpanel's analytics.js does—they use a 300ms delay: https://github.com/segmentio/analytics.js-core/blob/master/lib/analytics.js

(I'd extract it into some function in analytics.coffee)

PR #1418, opened 2018-08-15
Empty feed "Connect Twitter" CTA

broskoski:
Second part of #1383 

Still to do:

- [x] Mutate `has_seen_feed_connect_twitter` flag on modal 'Done'
- [x] Handle refresh to feed more gracefully
- [x] Open modal when callback dance is done
- [x] Loading / pagination on Twitter contact list

---

Comps:
screen shot 2018-08-15 at 7 36 48 pm
screen shot 2018-08-15 at 7 36 53 pm

---

Contact component:
screen shot 2018-08-15 at 7 38 13 pm


react/components/FollowButton/index.js
```diff
@@ -66,12 +66,14 @@ class FollowButton extends Component {
       ...rest
     } = this.props;
 
+    const is_followed = !!(followable && followable.is_followed);
```
dzucconi: What's this aboot?

broskoski: Forcing a boolean when the value doesn't exist. I will remove this, I just wanted to develop on Storybook first.

dzucconi: Ah weird yeah; it should never be nil. But I dunno what ur doing in Storybook

broskoski:
Alright @dzucconi this is essentially done, I just need to wire up the Authentication mutations, but I think that will take some coordination from you. We also need to figure out how to test this stupid auth dance without deploying to production. Thinking we should make an app and test app on the Are.na Twitter account that we can use locally.

broskoski:
OK getting closer. Still to do:
- [ ] Mutate `has_seen_feed_connect_twitter` flag on modal 'Done'
- [ ] Handle refresh to feed more gracefully

screen shot 2018-08-16 at 3 42 46 pm

broskoski:
follow-from-feed

react/components/ConnectTwitter/components/Contact/index.js
```diff
@@ -0,0 +1,84 @@
...
+      <Container>
+        <Identifier>
+          <UserAvatar user={user} mr={5} />
+          <Username>{user.name}</Username>
+        </Identifier>
+        <FollowButton
+          id={user.id}
+          type="USER"
+          followNode={<FollowText />}
+          unfollowNode={<UnfollowText />}
```
dzucconi: Why didn't we just pass a state prop to this component? I know we talked about this but I literally just did this for the dropdown and it's much cleaner.

dzucconi: https://github.com/aredotna/ervell/pull/1421/files#diff-7741a86c681d0a2c2fa8bda65f777d00 provideChildrenWithProps and pass along the follow/unfollow state

react/components/ConnectTwitter/index.js
```diff
@@ -0,0 +1,126 @@
...
+          query={connectTwitterQuery}
+          variables={{
+            per: 30,
+          }}
+        >
+          {({
+            loading, error, data, fetchMore,
+          }) => {
+            if (loading) return 'Loading...';
+            if (error) return `Error! ${error.message}`;
```
dzucconi: Lose the `Error!` ?

react/components/FollowButton/index.js
```diff
@@ -66,12 +73,14 @@ class FollowButton extends Component {
...
+
     return (
       <span onClick={this.toggleFollow} role="button" tabIndex={0} {...rest}>
         {{
-          true: 'Unfollow',
-          false: 'Follow',
-        }[followable.is_followed]}
+          false: followNode,
+          true: unfollowNode,
+        }[isFollowed]}
```
dzucconi: So this could just be {children} with a default prop of a component that displays Follow/Unfollow, which would then let you customize it accordingly.

broskoski: Ah got it, makes sense.

react/components/UI/GenericButton/index.js
```diff
@@ -92,4 +92,28 @@ export const GenericButton = styled.button`
...
+    border-top-color: ${x => x.theme.colors.gray.bold};
+    color: ${x => x.theme.colors.gray.bold};
+  }
+
+  &:active {
+    border-top-color: ${x => x.theme.colors.gray.active};
+    color: ${x => x.theme.colors.gray.active};
+  }
+`;
+
```
dzucconi: You don't have to do this but we need to separate these out into their own files then have the index export them all. (What input does)

react/pages/feed/EmptyConnectTwitter/index.js
```diff
@@ -0,0 +1,121 @@
...
+    const { updateFlag } = this.props;
+
+    updateFlag({
+      variables: {
+        name: 'HAS_SEEN_FEED_CONNECT_TWITTER',
+        value: true,
+      },
+    })
+      .then(() => { axios.get('/me/refresh'); })
+      .then(() => { setTimeout(() => { window.location = '/feed'; }, 400); });
```
dzucconi: Why the setTimeout?

broskoski: Sigh, to let the cache clear. I'm actually still having issues where occasionally after "Done" is pressed the page refreshes and this message is still there, so there has to be a better way.

dzucconi: Which cache?

broskoski: Passport / user cache.

dzucconi: So can we just fix that route then?

dzucconi: I really don't want setTimeouts littered around; there shouldn't be anything preventing us from OKing on the callback to clear whatever it's doing.

PR #1429, opened 2018-08-21
Add headers to settings app

broskoski:


apps/settings/routes.coffee
```diff
@@ -3,9 +3,13 @@
...
   policy.authenticate(user.get('access_token'))
   customer.authenticate(user.get('access_token'))
+  
+  headers = {
+    'X-AUTH-TOKEN': user.get('access_token'),
+  }
 
   Promise.all [
-    user.fetch()
+    user.fetch({ headers })
```
dzucconi: I don't think coffeescript does this?

broskoski: it does apparently

dzucconi: Ah wow you right

PR #1438, opened 2018-08-23
Remove useless conditional on block sidebar

broskoski:
This insane conditional was preventing blocks from loading in cases where:
- the block visibility was private
- the author was not the logged in user
- the block was text

All the conditional below covers the actual concerns that this stupid one was trying to cover for. I realize all this stuff is dead in the water but I ran into it twice and it was really annoying.


broskoski:
Note, the actual error that was preventing rendering was the MuteButton component attempting to mount on an element that was not present. The `mountWithApolloProvider` function DOES account for this like so:

```
if (!mountNode) return false;
```

But for some reason this stopped execution on `postRender`. Still haven't figured that part out.

PR #1444, opened 2018-08-27
Road to New Channel

dzucconi:
This gets the custom pulldown into manage channel

@csherron once this is merged poke at this in production; there's a few things missing but I'm wondering if it's good enough to actually use here.


dzucconi:
@csherron should we use the new colored-title-input in edit too?

react/components/ManageGroup/index.js
```diff
@@ -43,15 +43,19 @@ class ManageGroup extends Component {
...
   }
 
+  static getDerivedStateFromProps(nextProps) {
+    const { data: { loading } } = nextProps;
+    if (loading) return null;
+
+    const { data: { group: { name } } } = nextProps;
+    return { name };
+  }
+
```
dzucconi: This is one of those things that's going to come up a lot; there's a new static lifecycle method in React called `getDerivedStateFromProps` that is VERY useful here for figuring out state from the cached data.

PR #1449, opened 2018-08-29
Add react-safe-input and handle early input

### LINKED ISSUE #1443: Login Form on Firefox does not fire onChange events if login is autofilled.
## Steps to replicate
* Turn autocomplete on in Firefox
* Sign in
* Log out and sign back in again

## Expected result
If the element is prefilled, you should be signed in normally.

## Actual result
The page REFRESHES. No error's thrown, all required fields and validation ignored.

Nothing about the way Firefox handles forms in this state makes any sense. Even upon setting a `ref` for each form element and inspecting their values on `componentDidMount`, the `value` field was completely empty. EMPTY.

broskoski:
Also brings React up to date.

Closes #1443


dzucconi:
Perfect; I'd add a comment mentioning the github issue that tracks this in React.

PR #1463, opened 2018-08-31
Logs mount times

dzucconi:


dzucconi:
Hey @damassi have you encountered any low hanging fruit for slow compile times using babel-register? (We had to request a longer boot time on Heroku because of it)

damassi:
Yah that’s a bit weird, never encountered it🤔

PR #1464, opened 2018-09-04
Webpack 4

dzucconi:
I'm going to just work with this in development while I build out the user dropdown.

Build times went down but nothing *dramatic* though they seem to vary quite a bit.


models/block.coffee
```diff
@@ -1,17 +1,10 @@
-#
-# Model for a single block
-#
-
 Base = require "./base.coffee"
 sd = require("sharify").data
 _ = require 'underscore'
 _s = require 'underscore.string'
 moment = require 'moment'
-parseDomain = require 'parse-domain'
```
dzucconi: Sort of funny, this was probably the biggest dependency in the entire bundle and it was never used.

dzucconi:
Mixed bag now; including UserDropdown component (not in use yet). Going to merge and take a look on staging.

PR #1469, opened 2018-09-05
Deploy minor form fix

dzucconi:
Really just a minor change with a bunch of nonsense from the reverts.


dzucconi:
Side note; as I wait for the server with this config to boot up locally; I gotta say; Webpack 4 IS at least faster.

PR #1471, opened 2018-09-06
Webpack 4

dzucconi:


dzucconi:
https://www.styled-components.com/docs/faqs#why-am-i-getting-a-warning-about-several-instances-of-module-on-the-page

Trying to make sense of how to reproduce our original configuration using the runtimeChunk settings.
https://gist.github.com/sokra/1522d586b8e5c0f5072d7565c2bee693

This is some serious nonsense.

dzucconi:
OK! Finally got this working. We just have to share a single runtime (without specifying that Webpack embeds the runtime separately in each split package, which then initializes modules separately as well). So that's what was causing styled-components double initializing.

I'm also thinking we should pull out a separate vendor bundle. I'm going to post the analysis of each scenario.

dzucconi:
Single commons (which presumably would change pretty frequently)

webpack.config.js
```diff
@@ -120,17 +134,16 @@ const config = {
   },
 };
 
+if (ANALYZE_BUNDLE) {
+  config.plugins.push(new BundleAnalyzerPlugin());
```
damassi: This one is a real work of art

webpack.config.js
```diff
@@ -71,13 +73,28 @@ const config = {
...
+        // vendor: {
+        //   name: 'vendor',
+        //   chunks: 'initial',
+        //   test: /[\\/]node_modules[\\/]/,
+        // },
+        default: {
+          name: 'common',
+          chunks: 'all',
+          minChunks: 2,
+        },
```
damassi: Been poking at this on and off in Force but the bundle sizes are a bit larger for some reason. Need to return to it at some point.

PR #1495, opened 2018-09-19
Fix error in notifications feed

dzucconi:


dzucconi:
I don't think anything else reuses that template?

PR #1503, opened 2018-09-20
Fix FOUCs

dzucconi:
Closes https://github.com/aredotna/ervell/issues/1502


apps/feed/templates/feed.jade
```diff
@@ -3,18 +3,27 @@ extends ../../../components/layout/index--infinite
 block meta
   != homeComponent.styles
 
+  //- Not present on /notifications
```
broskoski: This should not be in meta

PR #1523, opened 2018-10-26
Blog

1aurabrown:
Fetches the following from contentful

- fetch a list of featured posts -> homepage blog module
- fetch a post by slug -> blog show
- fetch all posts -> blog index

Blog post body is a structure of nodes returned as json, using the 'Rich Text' field type of theirs which is currently in beta. Contentful provides an out-of-the-box parser which turns these abstract nodes into html. I extend their parser with a method for rendering assets (which is actually not defined by default in their renderer). this returns a <figure> tag similar to the one that was being generated by jekyll, for the sake of consistency, but includes an img with srcset urls and sizes.

Posts now have a linked author, whose name and bio can be filled out on contentful
Validations for fields on contentful can all be configured in their interface, Damon has a login. I can elaborate on this more offline.

I also simplified the CSS styles a bit. I skipped backbone and didn't use jade for the image rendering because it seemed convoluted, but I can do it that way if you prefer.


1aurabrown:
FYI, more on the rich text rendering:
https://www.npmjs.com/package/@contentful/rich-text-html-renderer
https://www.contentful.com/developers/docs/tutorials/general/getting-started-with-rich-text-field-type/

collections/posts.coffee
```diff
@@ -1,9 +1,35 @@
-{ Collection, Model } = require 'backbone'
-{ BLOG_URL } = require('sharify').data
+contentful = require "contentful"
```
broskoski: Don't feel beholden to writing coffee script. In fact, I think we prefer if all new stuff is straight ES6 js!

broskoski: Actually, strike that. It matters little.

broskoski:
Looks great, nice work!

@1aurabrown one practical concern here is that we will need to re-populate Contentful with the contents of the current blog before merging / deploying.

lib/setup.coffee
```diff
@@ -135,9 +135,16 @@ module.exports = (app) ->
 
   console.log 'Mounting apps...'
 
+  console.log 'Watching for changes...'
```
dzucconi: Not needed (prod isn't going to watch for changes)

yarn-error.log
```diff
@@ -0,0 +1,12396 @@
+Arguments:
```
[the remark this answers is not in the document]
1aurabrown: ok

lib/setup.coffee
1aurabrown: oh i see, i thought it was informative for the dev environment.

1aurabrown:
@dzucconi @broskoski updated, i will send out an email with some of my follow-up questions and a little how to on editing content.

@broskoski yes you're right content should be pre-populated before we flip the switch on this. I supposed that could be automated somehow but I'm not sure it saves any time.

broskoski:
@1aurabrown I'm going to close this PR, rebase with my version and resubmit!

p.s. WELL DONE 👏

PR #1527, opened 2018-11-06
Roughs in a basic RadioOptions component

dzucconi:
This is me avoiding the mega PR scenario. PR early and PR often.

radio

### DIFF
--- a/react/stories/RadioOptions.stories.js
+++ b/react/stories/RadioOptions.stories.js
@@ -1,0 +1,59 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import { action } from '@storybook/addon-actions';
+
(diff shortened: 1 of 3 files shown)


react/stories/RadioOptions.stories.js
```diff
@@ -0,0 +1,59 @@
...
+              <Text f={4} mb={3} selected={selected}>
+                <strong>Basic</strong>
+              </Text>
+
+              <Text f={1} color="gray.medium">
+                Basic members are limited to 100 private blocks.
+              </Text>
+            </div>
+          )}
+        </RadioOptions.Option>
```
dzucconi: Sort of similar to your Tab situation @broskoski

broskoski: Ah, interesting!

PR #1534, opened 2018-11-08
V/A Q/A (3)

dzucconi:
So: we move the contents stuff to be client-side only (which improves initial page load) and then convert the fetchPolicy to network-only; which fixes the infinite state issues.

### DIFF
--- a/react/components/CompactChannel/fragments/compactChannel.js
+++ b/react/components/CompactChannel/fragments/compactChannel.js
@@ -9,9 +9,13 @@ export default gql`
     title
     owner {
       ... on Group {
+        __typename
+        id
         name
       }
       ... on User {
+        __typename
+        id
         name
       }
     }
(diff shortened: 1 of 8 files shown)


broskoski:
Sigh, there will likely be SEO implications with this.

dzucconi:
... get on gchat

dzucconi:
tbh i think the seo fear is unfounded but ill try to work around anyway. i do think loading the initial page faster is a better idea however.

dzucconi:
d4ec5a53d2bf19ba640b757dea27ff5cacd97bbd is this crazy? Probably. Does it work? Yes.

Amazingly I'm still seeing the fucking Apollo parallel fetch bug.

dzucconi:
Though seeing it *way* less. It looks like it only happens when it's also concurrent with the networkPolicy update. I'm going to leave this here for now and maybe re-approach with a deeper bugfix to the underlying issue.

PR #1542, opened 2018-11-11
Disable 'create' channel button on submit

broskoski:


react/components/NewChannelForm/index.js
```diff
@@ -102,6 +106,7 @@ class NewChannelForm extends Component {
           error: 'Error',
         }[mode]}
         onDone={this.handleSubmit}
+        disabled={isDisabled}
```
dzucconi: Why not just pass the mode in?

dzucconi: Eh, I guess this makes sense.

dzucconi: `disabled={mode !== 'resting}` ?

Issue #1544, opened 2018-11-12
profile QA 


csherron:
- [ ] Random view on profile should be randomized with each refresh. (Currently when you refresh a profile, the blocks persist, no matter where you visit a profile from, logged out even. The Random sort mode should be random every single time with every refresh.)

- [ ] logged out profiles are missing the CTA banners "Follow User Name on Are.na" "Join" "Log In"

- [ ] Connect button is visible on logged out profiles

- [ ] Lightbox is not paginating - only loops within whatever has been loaded

- [ ] There's a delay of the "follow" button showing up on profiles and channel blocks, only displaying the em dash for a few seconds after hovering

- [ ] the 'following' page has that same block hover / refresh glitch after you've scrolled a page or two down

broskoski:
"Connect button is visible on logged out profiles" - we will leave this as-is, currently redirects to /join

broskoski:
Will take 1 and 2 (for now)

dzucconi:
The delay thing is the preflight latency I mentioned (it was more confusing to have erroneous follow states for as long as they take ATM)

PR #1546, opened 2018-11-12
Add LoggedOutCTA component

broskoski:
Addressing a single item in #1544

### DIFF
--- a/react/components/LoggedOutCTA/components/LoggedOutProfileContent/index.js
+++ b/react/components/LoggedOutCTA/components/LoggedOutProfileContent/index.js
@@ -2,7 +2,7 @@ import React, { Component } from 'react';
 import { propType } from 'graphql-anywhere';
 import userFragment from 'react/components/LoggedOutCTA/fragments/user';
 
-import { Copy, Button, Container } from 'react/components/LoggedOutCTA/components/UI';
+import { Copy, Button, Container, ButtonContainer } from 'react/components/LoggedOutCTA/components/UI';
 
 
 export default class LoggedOutProfileContent extends Component {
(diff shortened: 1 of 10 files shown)


react/components/LoggedOutCTA/components/UI/index.js
```diff
@@ -0,0 +1,29 @@
...
+  pr: 4,
+})``;
+
+const Button = styled(GenericButtonLink).attrs({
+  mx: 3,
+})`
+  ${constants.media.mobile`
+    padding-left: 0.5em;
+    padding-right: 0.5em;
+  `}
```
dzucconi: This is like fine (use scale values though). But if you want you can also just configure this with an array where each value in the breakpoints https://github.com/aredotna/ervell/blob/master/react/styles/constants.js#L4-L5

react/components/LoggedOutCTA/index.js
```diff
@@ -0,0 +1,38 @@
...
+const Container = styled(Box).attrs({
+  py: 6,
+  px: 4,
+})`
+  position: fixed;
+  width: 100%;
+  left: 0;
+  right: 0;
+  bottom: 0;
+  border-top: 3px solid ${x => x.theme.colors.gray.semiLight};
```
dzucconi: borderColor: 'gray.semiLight'

dzucconi: But I find this mixing of styles a bit odd in general.

react/components/LoggedOutCTA/components/UI/index.js
```diff
@@ -0,0 +1,29 @@
...
+
+`;
+
+const Container = styled.div`
+  display: flex;
+  align-content: center;
+  justify-content: center;
+`;
```
dzucconi: What's with the separate file for all of these things anyway. None of these components really require you to do a styled component custom for it. If that's easier for you then fine but it's much faster IMO to just inline these props on the components themselves `<Text f={5} py={3} pr={4}>`...

react/components/LoggedOutCTA/components/LoggedOutProfileContent/index.js
```diff
@@ -0,0 +1,23 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+
+import { Copy, Button, Container } from 'react/components/LoggedOutCTA/components/UI';
+
+
+export default class LoggedOutProfileContent extends Component {
+  static propTypes = {
+    user: PropTypes.node.isRequired,
```
dzucconi: User isn't a node? It should be a fragment presumably?

PR #1549, opened 2018-11-15
Pass seed in to profile contents and channels

broskoski:
Closes part of #1544 (random sort should be different on refresh)

### DIFF
--- a/react/components/ProfileContents/index.js
+++ b/react/components/ProfileContents/index.js
@@ -29,7 +29,6 @@ export default class ProfileContents extends Component {
     per: 12,
     hasMore: true,
     q: null,
-    seed: Date.now(),
   }
 
   shouldComponentUpdate(nextProps, nextState) {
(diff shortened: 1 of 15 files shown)


react/components/ProfileContents/index.js
```diff
@@ -29,6 +29,7 @@ export default class ProfileContents extends Component {
     per: 12,
     hasMore: true,
     q: null,
+    seed: Date.now(),
```
dzucconi: Won't the seed be different between ssr and the client-side paging?

broskoski: Ah yeah, good point. Hm, any ideas?

dzucconi: Set it as a cookie or push it into the query params so that it's at a routing level.

broskoski:
Closing this PR and starting fresh

PR #1562, opened 2018-11-21
Individual billing

dzucconi:
## TODO
- [x] Support lifetime plans
- [x] Initialize Stripe with env publishable key instead of hardcoded


react/components/Billing/components/BillingForm/index.js
```diff
@@ -0,0 +1,302 @@
...
+import CancelPremium from 'react/components/Billing/components/CancelPremium';
+import StatusOverlay from 'react/components/Billing/components/StatusOverlay';
+
+const OPERATIONS = {
+  ADD_NEW_CREDIT_CARD: 'ADD_NEW_CREDIT_CARD',
+  CHANGE_DEFAULT_CREDIT_CARD: 'CHANGE_DEFAULT_CREDIT_CARD',
+  CHANGE_PLAN_ID: 'CHANGE_PLAN_ID',
+  REENABLE_PREMIUM: 'REENABLE_PREMIUM',
+  CANCEL_PREMIUM: 'CANCEL_PREMIUM',
+};
```
dzucconi: Working on this now; trying to clean up the state. I hate that I'm realizing I basically need a state machine at this point.

csherron:
- [ ] some quick QA - haven't tried entering card info yet so this all happens before you activate anything:

- [ ] vertically center the loader 

- [ ] when not having signed up, the activate button should be greyed out until all necessary forms have been filled

- [ ] shouldn’t have cancel premium text until you’ve officially activated

- [ ] make top padding 48px (currently it’s at 20)

- [ ] small ‘billed monthly’ underneath the price is redundant doesn’t need to be there

- [ ] have a responsive breakpoint for the left/right padding where it eventually goes full-width if you squeeze the browser window enough

- [ ] the billing form shows up underneath the other tab sections (general, email notifs)

- [ ] the first credit card form field needs to be active when toggling a premium option

- [ ] remove “changes” label

- [ ] left-align the $$$ loader

csherron:
- [ ] can the list of cards be in the order of most recently added instead of the current card always floating to the top?

- [ ] after removing a card and being left with only one card - the state needs to return to it’s normal grey bar, currently it remains at the radio list with nothing else to select

- [ ] if already activated, when you click back to basic, hide the block count bar and blue text and put the ‘cancel premium’ button directly below the ‘basic’ row (above yearly premium)

- [ ] when adding a new card - the blue alert shouldn’t say ‘subscribed youre all set’ it should say ‘New card added - you’re all set!’ or something — and does it still need to be blue or could it be yellow? 

- [ ] move yellow grace period message to the very top

- [ ] a ‘billed to’ form appears underneath the yellow grace period message after refresh, but when you first cancel premium it’s not available. i’m not sure which one is a bug, but i think its good that users can remove their card after the fact (however you WILL need to be able to remove your card via a ‘remove card’ button - so maybe we need to add one after all)

coupons:

- [ ] 100% off for 3 months - my billing total doesn’t change, should it?

- [ ] the coupon messages are worded kind of strange - it should say “$5/45 off once” instead of 5/45 (USD)

- [ ] after entering a coupon and saving changes, then refreshing the page, the small “currently billing…” text remains unchanged, as if the coupon didn’t work

dzucconi:
For whatever reason the one of the builds fails here cuz of the branch I pushed to upstream. Ignore. @broskoski good to merge?

react/stories/RadioOptions.stories.js
```diff
@@ -13,47 +13,52 @@ storiesOf('RadioOptions', module)
...
+            <Text f={4} mb={3} selected={selected}>
+              <strong>Monthly Premium {JSON.stringify({ selected })}</strong>
+            </Text>
           )}
         </RadioOptions.Option>
+
+        <RadioOptions.Option value="lifetime" disabled>
+          {({ selected }) => (
+            <Text f={4} mb={3} selected={selected}>
+              <strong>Lifetime Premium (disabled) {JSON.stringify({ selected })}</strong>
```
broskoski: Should this say "(disabled)"?

broskoski:
Minor comment above, otherwise LGTM. Well done, its a pleasure to use.

react/stories/RadioOptions.stories.js
dzucconi: it's a story

broskoski: Sorry, missed that 👍

PR #1563, opened 2018-11-21
Fixes default value

dzucconi:
I personally find it weird that you can have GRAPHQL_ENDPOINT set to something like localhost but then this comes in. Unfortunately the way this is set up makes it semi-odd to set fallback to an alternate env variable. I'd rather it fail hard than fail mysteriously and work correctly for more cases.

### DIFF
--- a/config.coffee
+++ b/config.coffee
@@ -10,8 +10,8 @@ module.exports =
   PORT: 4000
   APP_URL: "http://localhost:4000"
   API_BASE: "http://api.are.na"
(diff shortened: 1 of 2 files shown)


dzucconi:
The more I look at this more I think it really is the case that what you want is not a separate URL because that could lead to insane behavior: you just want a proxied version of the underlying URL.

dzucconi:
I'll PR a different change that corrects this at some point.

PR #1574, opened 2018-11-28
Add user link to group-owned channels

broskoski:
screen shot 2018-11-28 at 4 54 59 pm


react/components/ChannelMetadata/components/ChannelMetadataInfo/fragments/channelMetadataInfo.js
```diff
@@ -10,5 +10,9 @@ export default gql`
     counts {
       followers
     }
+    user {
```
dzucconi: Add `__typename` and `id` as a general rule

PR #1576, opened 2018-11-29
Fix blog meta image and add description

broskoski:


components/contentful/helpers.coffee
```diff
@@ -5,7 +5,7 @@ cheerio = require 'cheerio'
 module.exports =
   metaImage: (image) ->
     return '' unless image.fields
-    return "#{image.fields.file.url}?w=1024"
+    return "https:#{image.fields.file.url}?w=1024"
```
dzucconi: Odd that they would require a protocol? Shrug

PR #1587, opened 2018-12-03
Re-enables Groups

dzucconi:
Not to be merged (yet). Just wanted to squash all of these into a single commit.


dzucconi:
Updated

PR #1591, opened 2018-12-04
Fixes applying coupons to fresh accounts

dzucconi:


react/components/Billing/components/BillingForm/index.js
```diff
@@ -158,7 +158,7 @@ class BillingForm extends PureComponent {
   handleCouponCode = (coupon_code) => {
     this.setState(prevState => ({
       coupon_code,
-      operations: this.addOperation(prevState.operations, 'APPLY_COUPON_CODE'),
+      operations: this[coupon_code === '' ? 'removeOperation' : 'addOperation'](prevState.operations, 'APPLY_COUPON_CODE'),
```
dzucconi: Fixes a bug where if you try to add an invalid coupon, then delete it, we were still trying to apply empty string which was causing a Stripe error

PR #1604, opened 2018-12-11
[Not for merging] React / Apollo QuickSearch Progress

broskoski:
Just getting this up so we can look at the code together @dzucconi .... like we both love doing.

screen shot 2018-12-10 at 8 27 56 pm


broskoski:
Nevermind, will open a clean version in the morning, this is a result of me checking out a new branch from a different groups branch.

PR #1613, opened 2018-12-12
Add inner shadow to top bar on group-related pages

broskoski:
screen shot 2018-12-12 at 3 01 00 pm
screen shot 2018-12-12 at 3 01 10 pm
screen shot 2018-12-12 at 3 01 34 pm

### DIFF
--- a/apps/channel/queries/channelCan.js
+++ b/apps/channel/queries/channelCan.js
@@ -4,6 +4,15 @@ export default gql`
   query channel_can($id: ID!) {
     channel(id: $id) {
       id
+      owner {
+        __typename
+        ... on Group {
+          id
+        }
+        ... on User {
+          id
+        }
+      }
       can {
         add_to
         update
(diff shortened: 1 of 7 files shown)


apps/channel/queries/channelCan.js
```diff
@@ -4,6 +4,15 @@ export default gql`
...
       id
+      owner {
+        __typename
+        ... on Group {
+          id
+        }
+        ... on User {
+          id
+        }
+      }
```
dzucconi: So this isn't related to channelCan and shouldn't be in here. Could just make this it's own fragment.

dzucconi: OR—just rename this since it's legacy

dzucconi: legacyChannelData or somethin

broskoski: I shall

Issue #1572, opened 2018-11-28
Groups UI


dzucconi:
- [ ] Refresh group on settings save
- [ ] Fix pulldown stacking in new channel modal (needs to be viewport aware fixed position)
- [ ] user pulldown needs to stop at bottom margin and scroll when it overflows
- [ ] persist close on group tip in new group modal
- [ ] Persist collapsed groups state in user dropdown
- [ ] Re-enable groups features

csherron:
GENERAL
- [ ] can the lock symbol appear on groups with only private channels at first? the lock then disappears as soon as public channels are created or transferred to that group
- [ ] index view right margin is not lining up / hanging off, could be something to do with the padding between columns?

HEADER
- [ ] ‘replace logo’ button is too small - just make normal size button and bleed it over the image area
- [ ] group header avatar needs to vertically flex and fit if the logo is not a square aspect ratio (no cropping or letterboxing) - for example if you uploaded a really wide and skinny logo, there would be less padding above and below the logo zone
- [ ] make group badge clickable with the title in both header and scrolled view top bar, (match hoverstates as well)

METADATA
- [ ] members paragraph list is too wide - needs to be 300px wide with 15px right margin (double check all paragraph widths in metadata)
- [ ] does the “started by” attribution carry over even after transferring a channel from a group to a different group or person? haven’t been able to test just wondering
- [ ] group/channel followers links needs to appear below, not above, the attribution text (admin / started by)
- [ ] group info admin name should be all on one line
- [ ] can ‘owned by’ sorting category be visible in channels view too? is ‘owned by’ the correct term, or can we sort by ‘created by’ with a list of authors and their channels?

TRANSFER
- [ ] if you own the group, transferring a channels’ ownership to that group should be instantaneous and not require any email confirmation
- [ ] group should show up in search results (okay if its square here)

NEW CHANNEL
- [ ] in author dropdown, get rid of (me) and (group) indicators and just use the small group badge next to group names

NEW GROUP
- [ ] put ‘once your group is created’ helper text inside the yellow banner
- [ ] new group yellow box copy - “secret group” isn’t a feature yet - try “You can also create a group to transfer and separate channels from your personal profile”

EDIT GROUP
- [ ] add/edit members section doesn’t need “invite” label - can be full width
[...]

PR #1620, opened 2018-12-14
More group QA

broskoski:
Picks off some more various items on https://github.com/aredotna/ervell/issues/1572

### DIFF
--- a/react/components/FollowButton/fragments/followable.js
+++ b/react/components/FollowButton/fragments/followable.js
@@ -17,5 +17,13 @@ export default gql`
         followers
       }
     }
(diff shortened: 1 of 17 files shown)


react/components/FollowButton/fragments/followable.js
```diff
@@ -17,5 +17,13 @@ export default gql`
...
       }
     }
+
+    ... on Group {
+      id
+      is_followed
+      counts {
+        followers
+      }
+    }
```
dzucconi: We really need to remove the counts from this

PR #1625, opened 2018-12-18
[Don't merge] Assign author in manage channel

broskoski:
I was going to PR this into the `reenablegroups-2` branch but there would be merge conflicts. I can also close this when ready and commit there or some other wild advanced git idea is fine too.

author


react/components/NewChannelForm/index.js
```diff
@@ -96,81 +99,131 @@ class NewChannelForm extends Component {
...
+                    borderless
+                    autoFocus
+                    required
+                    value={title}
+                    onChange={this.handleTitle}
+                    errorMessage={attributeErrors.title}
+                  />
+                </LabelledInput>
+
+                {counts.groups > 0 &&
```
dzucconi: This whole check for group existence seems overcomplicated and unnecessary. The author field can just handle this by itself and if there's no groups then it defaults to you and is disabled.

react/components/NewChannelForm/index.js
```diff
@@ -99,7 +99,7 @@ class NewChannelForm extends Component {
     const isDisabled = !(mode === 'resting' || mode === 'error');
 
     return (
-      <Query query={groupsCountQuery}>
+      <Query query={newChannelGroupsQuery}>
```
dzucconi: I'd just make this the newChannel query with a fragment for the authorship component in there.

broskoski:
Closing and re-pring to reenablegroups-2

PR #1631, opened 2018-12-20
More group QA

broskoski:
More things from #1572 

Namely, updates channel pulldown copy when author type changes, and updates the empty profile component to allow it to open the new channel modal (and pre-set the author) if the profile is the current users.


broskoski:
empty

react/components/NewChannelForm/index.js
```diff
@@ -18,17 +18,29 @@ class NewChannelForm extends Component {
...
+      mode: 'resting',
+      title: '',
+      description: '',
+      visibility: 'CLOSED',
+      group_id: props.group_id,
+      authorType: props.authorType,
+      visit_channel: true,
+      errorMessage: null,
+      attributeErrors: {},
+    };
```
dzucconi: If you want to set a value in state with a props you don't need to pull it into the constructor you can just use:

```javascript
state = {
  foo: this.props.foo,
}
```

react/components/NewChannelForm/index.js
```diff
@@ -52,8 +64,11 @@ class NewChannelForm extends Component {
     this.setState({ visibility });
 
   handleAuthor = (group_id) => {
-    if (group_id === 0) return;
-    this.setState({ group_id });
+    if (group_id === 0) {
```
dzucconi: I don't really recall this `group_id === 0` thing — it's a work around?

react/components/NewChannelForm/index.js
broskoski: ah, didn't know that

react/components/NewChannelForm/index.js
broskoski: Yeah, was there before. Looks like the mutation accepts a group_id rather than a `MemberInputType`.

react/components/NewChannelForm/index.js
broskoski: will fix

react/components/NewChannelForm/index.js
dzucconi: Ah I see — Weird I have no idea why I made the mutation like that.

PR #1642, opened 2019-01-04
More group Q/A - update user dropdown and group member list

broskoski:
screen shot 2019-01-04 at 10 39 31 am
screen shot 2019-01-04 at 10 39 47 am

### DIFF
--- a/react/components/UserDropdown/components/MyGroups/fragments/myGroups.js
+++ b/react/components/UserDropdown/components/MyGroups/fragments/myGroups.js
@@ -6,7 +6,7 @@ export default gql`
   fragment MyGroups on Me {
     __typename
     id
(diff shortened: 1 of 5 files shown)


react/components/UserDropdown/components/MyGroups/fragments/myGroups.js
```diff
@@ -6,7 +6,7 @@ export default gql`
   fragment MyGroups on Me {
     __typename
     id
-    is_my_groups_dropdown_visible: flag(name: "is_my_groups_dropdown_visible")
+    is_my_groups_dropdown_hidden: flag(name: "is_my_groups_dropdown_hidden")
```
broskoski: I know this is slightly weird, but we have falsey nils by default and this needs to be open by default.

dzucconi: 👍

PR #1648, opened 2019-01-07
Opts-in to cancel

dzucconi:


dzucconi:
Apparently the only bit that went un-pushed (thankfully)

PR #1679, opened 2019-01-11
Removes Airbrake

dzucconi:


dzucconi:
Disabling Airbrake for now. Apparently this got fixed (https://github.com/airbrake/airbrake-js/issues/508) but we currently aren't using it anyway. Will likely revert in the future.

PR #1686, opened 2019-01-11
More error handling

dzucconi:


apps/feed/index.js
```diff
@@ -72,7 +72,19 @@ const renderFeed = (req, res, next) => {
 
       return res.render('feed');
     })
-    .catch(next);
+    .catch((err) => {
+      const STATUS_CODE = err.graphQLErrors[0].extensions.code;
```
dzucconi: Actually I probably have to be more defensive here, annoyingly.

broskoski: Ya :/

PR #1687, opened 2019-01-11
Adds New Relic

dzucconi:


dzucconi:
We just need to add the New Relic add on before we deploy (though if there's no license key it doesn't matter will just ignore.)

Issue #1693, opened 2019-01-14
Next and Prev buttons not working for Lightbox in feed view


broskoski:
## Steps to replicate
* Hit next or previous in lightbox mode

## Expected result
You should go to the next block

## Actual result
You do not

PR #1696, opened 2019-01-15
Fix Lightbox prev / next on feed

broskoski:
Closes https://github.com/aredotna/ervell/issues/1693

This doesn't automatically load the next page when the user reaches the end (almost), but at least next / previous work now.

### DIFF
--- a/react/components/Feed/index.js
+++ b/react/components/Feed/index.js
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import { Query } from 'react-apollo';
 import styled from 'styled-components';
 import InfiniteScroll from 'react-infinite-scroller';
-import { map, flatten, filter } from 'underscore';
+import { map, flatten } from 'underscore';
 
 import feedQuery from 'react/components/Feed/queries/feed';
 
@@ -39,8 +39,7 @@ export default class Feed extends PureComponent {
     if (!data) return [];
 
     const { me: { feed: { groups } } } = data;
-    const objects = flatten(map(groups, group => group.objects));
-    return filter(objects, object => (object.__typename !== 'Channel' && object.__typename !== 'Group' && object.__typename !== 'User'));
+    return flatten(map(groups, group => group.objects));
   }
 
   render() {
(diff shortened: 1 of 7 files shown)


react/util/updateMediatorBlocks.js
```diff
@@ -0,0 +1,11 @@
...
+import Blocks from 'collections/blocks.coffee';
+
+export default ({ blocks }) => {
+  mediator.shared.blocks = new Blocks(blocks.map(k => ({
+    id: k.id,
+    class: k.__typename,
+    base_class: 'Block',
+    unloaded: true,
+  })));
+};
```
dzucconi: Just wondering why this is needed?

react/components/Feed/index.js
```diff
@@ -33,6 +35,14 @@ export default class Feed extends PureComponent {
     hasMore: true,
   }
 
+  getContext = (data) => {
+    if (!data) return [];
+
+    const { me: { feed: { groups } } } = data;
+    const objects = flatten(map(groups, group => group.objects));
+    return filter(objects, object => (object.__typename !== 'Channel' && object.__typename !== 'Group' && object.__typename !== 'User'));
```
dzucconi: https://github.com/aredotna/ervell/blob/master/react/components/Cell/components/Konnectable/index.js#L99 also?

react/components/Feed/index.js
```diff
@@ -63,10 +73,11 @@ export default class Feed extends PureComponent {
           return (
             <InfiniteScroll
               pageStart={1}
-              threshhold={500}
+              threshhold={1000}
               loader={<BlocksLoadingIndicator key="loading" />}
               hasMore={hasMore}
               loadMore={() => {
+                console.log('LOAD MORE');
```
dzucconi: 👋

react/util/updateMediatorBlocks.js
broskoski: I can remove it, but eventually the next page will load and `mediator.blocks` will need to be updated.

react/components/Feed/index.js
broskoski: Groups and users are on the feed, didn't seem like something that belonged in Konnectable?

dzucconi: I don't know I don't remember any of how I got it working. Realistically it doesn't belong in either place

PR #1703, opened 2019-01-16
Set channels view as default view and adds count to row

broskoski:
screenshot 2019-01-15 20 16 35


react/hocs/WithBlocksGridCount/index.js
```diff
@@ -0,0 +1,28 @@
...
+      return (
+        <SizeMe>
+          {({ size }) => {
+            const divider = parseInt(constants.blockAndGutter, 10);
+            const blocksGridCount = Math.floor(size.width / divider);
+            return (
+              <WrappedComponent blocksGridCount={blocksGridCount} {...rest} />
+            );
+          }}
+        </SizeMe>
```
dzucconi: Why is any of this necessary?

broskoski: The block count on the bottom right is dependent on how many blocks are visible on the screen.

dzucconi: This seems like a super over kill way of going about it. Couldn't you just check the el offsets?

dzucconi: What does this do on server vs client-side render?

dzucconi: Because presumably no matter what it's going to be wrong on initial render. Shouldn't we just design around this?

broskoski: Alright agreed. Will confer with Chris

PR #1707, opened 2019-01-16
RSS performance updates

### LINKED ISSUE #1702: RSS tweaks
* Explicitly error on private channels instead of rendering undefineds
* Cache public channels

broskoski:
Closes #1702

### DIFF
--- a/apps/rss/routes.coffee
+++ b/apps/rss/routes.coffee
@@ -17,8 +17,8 @@ posts = require "../../collections/posts.coffee"
     per: 25
 
   Q.allSettled([
-    channel.fetch()
-    blocks.fetch()
+    channel.fetch(cache: true)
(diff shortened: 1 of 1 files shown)


apps/rss/routes.coffee
```diff
@@ -34,10 +34,11 @@ posts = require "../../collections/posts.coffee"
     per: 25
 
   Promise.all([
-    author.fetch()
-    blocks.fetch()
+    author.fetch(cache: true)
+    blocks.fetch(cache: true)
```
dzucconi: What's the default TTL for this stuff?

broskoski: 1 hour

dzucconi: 👍

PR #1725, opened 2019-01-18
Sets gc flag for Node based on 2x Heroku dyno

dzucconi:
Re: https://devcenter.heroku.com/articles/node-memory-use#tuning-the-garbage-collector

### DIFF
--- a/react/util/compactObject.js
+++ b/react/util/compactObject.js
@@ -1,7 +1,14 @@
-export default (x = {}) =>
+export const isBlank = value =>
+  value === null || value === undefined;
+
+export const isNull = value =>
+  value === null;
+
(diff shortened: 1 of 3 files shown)


react/util/compactObject.js
```diff
@@ -1,7 +1,14 @@
...
+export const isNull = value =>
+  value === null;
+
+export default (x = {}, isBlankFn = isNull) =>
   Object.keys(x).reduce((memo, key) => {
-    if (x[key] !== null) {
+    if (!isBlankFn(x[key])) {
       return { ...memo, [key]: x[key] };
     }
+
```
dzucconi: Unrelated from another PR I'm working on. No change in behavior here.

PR #1728, opened 2019-01-18
Upgrades various dependencies, removes unused dependencies

dzucconi:


broskoski:
It doesn't appear that we are currently using the things you removed, LGTM.

PR #1733, opened 2019-01-21
Remove the need for getting profile counts

broskoski:
screenshot 2019-01-21 15 57 56
screenshot 2019-01-21 15 57 35
screenshot 2019-01-21 15 59 26


react/pages/profile/ProfilePage/components/EmptyMessageOrComponent/fragments/emptyOrTips.js
```diff
@@ -7,10 +7,6 @@ export default gql`
       id
       created_at(format: "%Y-%m-%dT%H:%M:%S.%L%z")
       is_me
-      counts {
-        blocks
-        channels
-      }
```
dzucconi: No, these look like they belong here?

react/pages/profile/ProfilePage/components/ProfileViews/index.js
```diff
@@ -55,7 +55,7 @@ const Channels = ({
 }) => (
   <EmptyMessageOrComponent
     identifiable={identifiable}
-    count={identifiable.counts.channels}
+    count={identifiable.counts.channels - 1}
```
dzucconi: Why are we subtracting 1 again?

broskoski: Sorry, leftover, will remove.

dzucconi: I guess this component can just handle this instead of it being a prop, also?

react/pages/profile/ProfilePage/components/EmptyMessageOrComponent/fragments/emptyOrTips.js
broskoski: They no longer do.

dzucconi: Well sorta, they would need to either be in a fragment for `ProfileViews` or rather just be handled within this component.

dzucconi: It's unclear what they belong to here

broskoski: Ah, they exist on the fragment above. Shoot, guess this is more just cleaning up the logic here (rather than removing the need for a count).

PR #1737, opened 2019-01-22
Full SSR pages (applies to authentication app)

dzucconi:


react/pages/authentication/ResetPasswordPage/index.js
```diff
@@ -18,13 +20,23 @@ export default class ResetPasswordPage extends Component {
...
+                <LoadingIndicator />
+              </CenteringBox>
+            );
+          }
 
           if (error) {
             return (
               <CenteringBox p={7} flexDirection="column">
+                <Head>
+                  <title>Sorry</title>
```
broskoski: Are.na / Sorry

dzucconi: That's what it renders

react/components/UI/Head/index.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Helmet from 'react-helmet-async';
+
+const Head = ({ children, ...rest }) => (
+  <Helmet defaultTitle="Are.na" titleTemplate="Are.na / %s" {...rest}>
```
dzucconi: @broskoski Template here

react/pages/authentication/ResetPasswordPage/index.js
broskoski: I know, it's funny

dzucconi: Are.na is sorry

react/components/UI/Page/index.js
```diff
@@ -0,0 +1,100 @@
...
+
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const GOOGLE_ANALYTICS_TRACKING_TAG = `
+  (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
+  (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
+  m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
+  })(window,document,'script','//www.google-analytics.com/analytics.js','ga');
+`;
```
dzucconi: Ah this probably needs to set the tracking ID?

dzucconi: Hm, this is annoying. Let's chat real quick.

dzucconi:
So lil thing I noticed that we'll have to figure out how to deal with: we should probably stop using/come up with a different method of handling sharify data that's set within routes. We're kind of using it in a  to just shovel data around globally, which is breaking on client-side route changes. You can see this in production by loading /sign_up server-side, then navigating to /log_in. You'll be redirected to /welcome on login.

PR #1739, opened 2019-01-23
React / Apollo explore page

broskoski:
screen shot 2019-01-23 at 12 14 04 pm


apps/index.coffee
```diff
@@ -12,6 +12,11 @@ console.time('feed')
 app.use(require('../apps/feed'))
 console.timeEnd('feed')
 
+
+console.time('explore')
+app.use(require('../apps/explore'))
+console.timeEnd('explore')
```
dzucconi: Why a separate app?

broskoski: Less confusing, it's not at all the same thing as explore and I always forget where it is.

dzucconi: I'd much rather us start to merge apps where it makes sense. It takes time to boot and this way they share a package. So similar things should definitely be grouped.

react/components/UI/CookieLinkUnlessCurrent/index.js
```diff
@@ -0,0 +1,51 @@
...
+
+const Link = styled(NavLink)`
+  display: block;
+  color: ${x => x.theme.colors.gray.regular};
+
+  ${({ isActive, theme }) => isActive && isActive() && `
+    color: ${theme.colors.gray.semiBold};
+    cursor: default;
+  `}
+`;
```
dzucconi: I think having this styling here for something this abstract is a problem. And also if you're abstracting some pattern somewhere, use it where you found the original.

dzucconi:
I should say: I'd rather have the copy paste over an awkward abstraction.

PR #1748, opened 2019-01-29
Do not render /explore contents server side

broskoski:


dzucconi:
👍

dzucconi:
Do this for search also

PR #1749, opened 2019-01-30
Set q to null

broskoski:


dzucconi:
While this is fine to get a deploy out—this actually does need a real fix. Will do.

PR #1751, opened 2019-01-30
Use new `channels` node for queries that don't involve search

broskoski:


broskoski:
@dzucconi I like your avatar

dzucconi:
IT'S ME

PR #1753, opened 2019-01-31
TopBar implementation

dzucconi:
Will hand over to @csherron for Q/A inside of storybooks before putting this in anywhere. Is fine to merge for now/not being used yet.


react/components/TopBar/components/PrimarySearch/components/PrimarySearchResults/index.js
```diff
@@ -0,0 +1,96 @@
...
+            );
+          }
+
+          const { results } = searches;
+          const selected = cursor && mod(cursor, results.length + 1);
+
+          this.selectResult(results[selected]);
+
+          return (
+            <React.Fragment>
```
broskoski: Interestingggg

dzucconi: what's interesting?

react/components/UI/SearchInput/index.js
```diff
@@ -128,6 +142,10 @@ class SearchInput extends Component {
           onBlur={onBlur}
           onChange={this.handleChange}
           defaultValue={query}
+          autoCorrect="off"
```
broskoski: Thank you

react/components/TopBar/components/PrimarySearch/components/PrimarySearchResults/index.js
broskoski: Have not seen this before

dzucconi: Ah, it's just so you can return a single child without an extra DOM element, like wrapping something in a <div>

PR #1755, opened 2019-02-01
TopBar + Profile

dzucconi:
### TODO

- [x] Remaining meta tags for profile (descriptions, images, etc)
- [x] Perfectly match existing top bar styling
- [x] Render new TopBar over old one on all other pages
- [x] Bind lightbox keyboard shortcuts on profile page
- [x] `ssr={true}` on profile contents for 🤖 
- [x] Ensure group channels are also greybar'd


react/apollo/index.js
```diff
@@ -66,11 +67,20 @@ export const initApolloClient = ({
...
+      },
     },
     resolvers: {
       Cookies: {
         get: (_obj, args) => cookies[args.name] || null,
       },
+      Sharify: {
+        get: (_obj, args) => {
+          const value = sharifyData[args.name];
+          return value === undefined ? null : value;
```
dzucconi: Probably should have just bit the bullet and did this earlier. Now we just access sharify through a client-side query.

Admittedly a bit weird but basically the issue is that importing it doesn't give you access to any data you might have set in the request locals. You, @broskoski, happened to just pull from an earlier SHA, hence my confusion.

broskoski: ah got it

broskoski: and yes, good idea

PR #1795, opened 2019-02-12
Supports dark mode

dzucconi:
Trying to minimize dependencies in any of the new code.

We don't want to rely on all.css, jQuery, Backbone, etc for this dumb feature. (We can delete all.css in the new pages as soon as we do block lightboxes, fwiw)


broskoski:
Beautiful

PR #1818, opened 2019-02-18
Add link to video tutorial for newly joined users

broskoski:


react/components/UserDropdown/components/SecondaryLinks/index.js
```diff
@@ -7,17 +7,25 @@ import SignOutLink from 'react/components/UserDropdown/components/SecondaryLinks
 export default class SecondaryLinks extends Component {
   static propTypes = {
     isPremium: PropTypes.bool,
+    recentlyJoined: PropTypes.bool,
```
dzucconi: `isRecentlyJoined`

react/components/UserDropdown/index.js
```diff
@@ -69,6 +70,8 @@ export default class UserDropdown extends Component {
 
           const { me } = data;
 
+          const recentlyJoined = moment().diff(moment(me.created_at), 'days') < 7;
```
dzucconi: This is fine for now, but in general I want to get rid of moment (https://github.com/you-dont-need/You-Dont-Need-Momentjs). I think this might be the only place in js that's importing it?

PR #1829, opened 2019-02-22
Random channel title

broskoski:
Feel free to suggest alternatives or merge and change it completely. This is a collaboration.

random-channel


react/components/NewChannelForm/index.js
```diff
@@ -14,6 +15,16 @@ import ChannelVisibilityPulldown from 'react/components/ChannelVisibilityPulldow
 import NewChannelGroups from 'react/components/NewChannelForm/components/NewChannelGroups';
 import LoadingIndicator from 'react/components/UI/LoadingIndicator';
 
+import RandomChannelIcon from 'react/components/NewChannelForm/components/MagicHat';
+
+import { random } from 'faker';
```
dzucconi: So: can you run a production build with ANALYZE_BUNDLE turned on and compare before and after this dependency. I know this has a lot of locales and I'm not sure what the filesize is here. (Hence me wanting to do this on the API side)

broskoski: Ah kk

PR #1838, opened 2019-02-26
Add autofocus to login and sign up forms (closes #1801)

### LINKED ISSUE #1801: no auto-select on join / sign in forms
currently the first form doesn't auto-select when you open the join form or log in form

broskoski:


react/components/LoginForm/index.js
```diff
@@ -136,6 +136,7 @@ export default class LoginForm extends Component {
           value={email}
           hasError={mode === 'error'}
           required
+          autofocus="true"
```
dzucconi: it's just `autoFocus`

broskoski: kk

Issue #1799, opened 2019-02-14
channel preview


csherron:
- [x] replace follow button on channel blocks with 'Preview'
<img width="387" alt="screen shot 2019-02-14 at 8 46 32 am" src="https://user-images.githubusercontent.com/3780330/52790983-e252fc00-3035-11e9-8410-d43c32c4ba37.png">

- [x] when clicking 'Preview', the channel text disappears and the bottom buttons are replaced by a single full-width 'Stop Preview' button. the standard throbber is the color of the channel, while blocks load in the background
<img width="377" alt="screen shot 2019-02-14 at 8 46 36 am" src="https://user-images.githubusercontent.com/3780330/52790990-e717b000-3035-11e9-91ad-eb35be51a561.png">

- [x] once 5-10 thumbnails have loaded, display a looping slideshow of those blocks, with each block appearing for half a second, no dissolve effects
<img width="537" alt="screen shot 2019-02-14 at 8 46 51 am" src="https://user-images.githubusercontent.com/3780330/52791008-f139ae80-3035-11e9-9f94-9dde9e828c0e.png">
<img width="378" alt="screen shot 2019-02-14 at 8 46 42 am" src="https://user-images.githubusercontent.com/3780330/52790997-ebdc6400-3035-11e9-99ce-b0f9265404f4.png">
<img width="366" alt="screen shot 2019-02-14 at 8 46 46 am" src="https://user-images.githubusercontent.com/3780330/52791000-ee3ebe00-3035-11e9-829b-0468a712fa9d.png">
<img width="354" alt="screen shot 2019-02-14 at 8 46 57 am" src="https://user-images.githubusercontent.com/3780330/52791015-f3037200-3035-11e9-8f9d-fd0b03f27811.png">

- [x] mousing over the slideshow will lower the image opacity to 10% and reveal the channel credits again, clicking the slideshow will take you to the channel, clicking 'stop preview' will return the channel block to it's normal state 
<img width="353" alt="screen shot 2019-02-14 at 8 47 17 am" src="https://user-images.githubusercontent.com/3780330/52791026-f72f8f80-3035-11e9-895e-fda50fb1e441.png">

PR #1854, opened 2019-03-06
ChannelPreview

dzucconi:
Closes https://github.com/aredotna/ervell/issues/1799


broskoski:
Looks extremely clean. WOW.

PR #1872, opened 2019-03-11
Roadmap

broskoski:
We are doing this quick and dirty to test with a very small amount of people first.

roadmap

### DIFF
--- a/react/pages/about/RoadmapPage/index.js
+++ b/react/pages/about/RoadmapPage/index.js
@@ -1,0 +1,234 @@
+import React, { PureComponent } from 'react';
+import styled from 'styled-components';
+import PropTypes from 'prop-types';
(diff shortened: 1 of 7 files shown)


react/pages/about/RoadmapPage/index.js
```diff
@@ -0,0 +1,234 @@
...
+
+export default class RoadmapPage extends PureComponent {
+  static propTypes = {
+    roadmap: PropTypes.shape.isRequired,
+  }
+
+  render() {
+    const { roadmap } = this.props;
+
+    console.log('roadmap', roadmap);
```
dzucconi: hi

broskoski: hi

Issue #1875, opened 2019-03-11
lock icon tweaks


csherron:
- [x] feed lock icon needs to be right of timestamp, not left
- [x] in quicksearch results, move lock to next to channel titles with 6px left padding
- [x] 'select privacy' dropdown - 2 more px left padding to lock icon
- [x] lock not showing up in channel bars on lightbox or index
- [x] add 2 more px left padding in notifications lock
- [x] add 2px more left padding on lock icons in user dropdown

## Punt
- [x] lock not showing up in channel bars on lightbox (will handle with new lightbox)
- [ ] i'm not seeing lock icon on connect block interface whilst inside channels

PR #1882, opened 2019-03-12
Lock positioning tweaks

dzucconi:
The majority of https://github.com/aredotna/ervell/issues/1875

### DIFF
--- a/react/components/UI/BorderedLock/index.js
+++ b/react/components/UI/BorderedLock/index.js
@@ -22,10 +22,10 @@ const Container = styled(Box).attrs({
   line-height: 0;
 `;
 
-const LockIconWithBorder = ({ ...rest }) => (
+const BorderedLock = ({ ...rest }) => (
   <Container {...rest} >
     <Lock />
   </Container>
 );
 
-export default LockIconWithBorder;
+export default BorderedLock;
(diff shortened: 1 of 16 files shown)


react/components/UI/BorderedLock/index.js
```diff
@@ -1,30 +1,31 @@
...
-
 import Icon from 'react/components/UI/Icons';
 import Box from 'react/components/UI/Box';
 
 // TODO: All this should be configurable if we need a similar style of icon
 const Lock = styled(Icon).attrs({
   name: 'Lock',
   color: 'gray.medium',
-  size: '10px',
+  size: '0.625rem',
```
dzucconi: @broskoski if you need to convert to rem or em depending on the context just divide the number u want by 16.

dzucconi: em when you want a value to scale based on the surrounding immediate context (typically padding/margins), rem when you want to scale based on the root font size (which shifts down on mobile)

react/components/UI/BorderedLock/index.js
```diff
@@ -1,30 +1,31 @@
...
 })`
 `;
 
-const Container = styled(Box)`
-  border: 1px solid ${translucentGray('regular')} !important;
+const Container = styled(Box).attrs({
+  border: '1px solid',
```
dzucconi: borders are sort of the only thing we use px for really

PR #1888, opened 2019-03-12
Fixes lock on connection lists

dzucconi:
Cherry picking from my current branch to fix this.


react/components/ConnectionSelectionList/components/SelectableChannel/index.js
```diff
@@ -122,15 +125,6 @@ export default class SelectableChannel extends Component {
...
     isSelected: false,
-    isHovered: false,
-  }
-
-  handleMouseEnter = () => {
-    this.setState({ isHovered: true });
-  }
-
-  handleMouseLeave = () => {
-    this.setState({ isHovered: false });
```
dzucconi: Using state for hovers is a hack around a very specific problem with Cells

PR #1894, opened 2019-03-14
Adds support for Recaptcha

dzucconi:


dzucconi:
Setting some env variables and getting this working on staging.

dzucconi:
Tho we skip it there fwiw. Maybe we shouldn't?

dzucconi:
Actually we def shouldn't since it could potentially be abused there too.

PR #1907, opened 2019-03-20
Adds ability to select premium plan on signup

broskoski:
signup

### DIFF
--- a/react/components/Billing/components/BillingForm/index.js
+++ b/react/components/Billing/components/BillingForm/index.js
@@ -39,9 +39,15 @@ class BillingForm extends PureComponent {
   static propTypes = {
     subscribeToPremium: PropTypes.func.isRequired,
     cancelPremiumSubscription: PropTypes.func.isRequired,
-
     applyCouponToSubscription: PropTypes.func.isRequired,
+    onSuccess: PropTypes.func,
     me: propType(billingFormFragment).isRequired,
+    plan_id: PropTypes.string,
+  }
+
+  static defaultProps = {
+    plan_id: null,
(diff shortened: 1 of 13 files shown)


react/components/Billing/components/BillingForm/index.js
```diff
@@ -322,7 +340,7 @@ class BillingForm extends PureComponent {
                 onClick={this.handleSubmit}
                 disabled={(
                   (!customer.default_credit_card) ||
-                  (operations.length === 0)
+                  (operations.length === 0 && !this.props.plan_id)
```
dzucconi: I'm not sure I follow the reasoning here? Can you explain?

react/components/Billing/components/BillingForm/index.js
```diff
@@ -39,20 +39,32 @@ class BillingForm extends PureComponent {
...
+    onSuccess: () => null,
   }
 
   state = {
     mode: 'resting',
     errorMessage: null,
     operations: [],
 
-    plan_id: this.props.me.customer.plan.id,
+    plan_id: this.props.plan_id || this.props.me.customer.plan.id,
```
dzucconi: I don't think this needs to be here if you are calling handlePlan in the `componentDidMount`?

react/components/Billing/components/BillingForm/index.js
broskoski: Taking this out

react/components/Billing/components/BillingForm/index.js
broskoski: And this too

PR #1911, opened 2019-03-22
Move feed to fully react rendered

broskoski:
Death to the old ways


broskoski:
Going to merge feed and explore apps in next PR...

PR #1912, opened 2019-03-22
Combine explore and feed apps

broskoski:


apps/feed/index.js
```diff
@@ -23,17 +24,21 @@ const render = (req, res, next) => {
...
     .catch((err) => {
-      const STATUS_CODE = getFirstStatusCode(err);
-
-      if (STATUS_CODE === 'UNAUTHORIZED') {
-        // This typically happens if the serialized user is "bad"
-        // or not actually logged in. If so: logout, then redirect somewhere.
-        // Falling through by using `next()` doesn't seem to actually purge the session.
-        req.logout();
-        return res.redirect('/log_in');
-      }
```
dzucconi: does this still work?

broskoski: How can I replicate what this is solving for? I didn't actually mean to take this out completely, but when I refactored, I was getting a redirect to `/log_in` when not logged in (instead of showing the homepage).

dzucconi: well that's not good. well first of all presumably this still redirects to the homepage when you're logged out? if that works then you could replicate this i believe by starting in staging: logging in; killing the server and restarting in production (this will give u bad auth which should simulate it). it's hard to debug admittedly.

broskoski: Ok, put it back in and realized it was `ensureLoggedInMiddleware` that was getting me.

dzucconi:
👍

PR #1917, opened 2019-03-25
Add home path middleware back

broskoski:


broskoski:
@dzucconi sorry accidentally pulled the trigger here. With the last deploy I broke 'home path settings' and this fixes it.

PR #1924, opened 2019-03-28
Remaining lightbox items

dzucconi:
This wraps up the core functionality for the lightbox. I've got a branch that enables this so I'll open a PR with that next and we'll Q/A this a bit.


react/components/UI/MentionTextarea/queries/userSuggestions.js
```diff
@@ -0,0 +1,18 @@
...
+
+export default gql`
+  query MentionTextareaUserSuggestions($query: String!) {
+    suggestions: searches {
+      users: collaborators(query: $query, limit: 4, types: USER) {
+        ... on User {
+          id: slug
+          label: name
+        }
+        ... on Group {
```
broskoski: I don't think anything is set up to handle group mentions (though it's a good idea).

dzucconi: Oh I know; I can delete this; this was to prevent the storybook from crashing because of the way mocks returns this.

PR #1931, opened 2019-04-01
Remove the whole complicated random seed setting

broskoski:
If you love randomness, you're going to love this PR.

(Note: we don't need the param setting bc we are rendering on the client anyway)


broskoski:
Screenshot 2019-04-01 11 13 27

PR #1939, opened 2019-04-01
Lightbox

dzucconi:


react/components/UI/Text/index.js
```diff
@@ -98,6 +103,12 @@ export const mixin = css`
   u {
     text-decoration: underline;
   }
+
+  ol,
+  ul {
+    list-style: disc;
+    margin-left: ${props => props.theme.space[5]};
+  }
```
dzucconi: FWIW: We don't support rendering nested lists in Markdown (currently). The old behavior was due to the client-side rendering which used a completely different lib. You can open an issue for that in the API if needed.

PR #1953, opened 2019-04-04
Port find friends tab to new component

broskoski:
Seems that the old one no longer works...as reported by @okwme.


broskoski:
Hi @okwme

PR #1956, opened 2019-04-05
Reset document title when block lightbox closes

### LINKED ISSUE #1933: Revert title back to channel or profile after lightbox is closed
## Steps to replicate
* Open a block within a channel or profile
* Close it

## Expected result
Title (meta) should go back to channel or profile title

## Actual result
It does not

broskoski:
Closes #1933

### DIFF
--- a/react/components/BlockLightbox/index.js
+++ b/react/components/BlockLightbox/index.js
@@ -25,14 +25,12 @@ export default class BlockLightbox extends PureComponent {
     context: PropTypes.oneOf(['MODAL', 'PAGE']),
     layout: PropTypes.oneOf(['DEFAULT', 'FULLSCREEN']),
     children: PropTypes.node,
-    onClose: PropTypes.func,
   }
 
   static defaultProps = {
     layout: 'DEFAULT',
     context: 'PAGE',
     children: null,
-    onClose: () => null,
   }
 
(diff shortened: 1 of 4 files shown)


react/components/Cell/components/Konnectable/index.js
```diff
@@ -58,11 +58,13 @@ export default class Konnectable extends PureComponent {
       id: PropTypes.number.isRequired,
     })),
     isPreviewable: PropTypes.bool,
+    documentTitle: PropTypes.string,
   }
 
   static defaultProps = {
     context: [],
     isPreviewable: true,
+    documentTitle: document.title,
```
dzucconi: We can get around not putting this here.

dzucconi: You can pass through the `onClose` for instance.

react/components/BlockLightbox/index.js
```diff
@@ -57,6 +60,7 @@ export default class BlockLightbox extends PureComponent {
         {layout === 'DEFAULT' &&
           <BlockLightboxMetadataPane
             block={block}
+            onClose={onClose}
```
dzucconi: Why does the metadata pane get this prop?

PR #1962, opened 2019-04-09
Channel loading

dzucconi:


react/components/ChannelContents/components/ChannelContentsPage/index.js
```diff
@@ -0,0 +1,102 @@
...
+import Waypoint from 'react-waypoint';
+import { Query } from 'react-apollo';
+
+import channelContentsPageQuery from 'react/components/ChannelContents/components/ChannelContentsPage/queries/channelContentsPage';
+
+import Cell from 'react/components/Cell';
+import GridItem from 'react/components/UI/Grid/components/GridItem';
+import ChannelContentsPageSkeleton from 'react/components/ChannelContents/components/ChannelContentsPageSkeleton';
+
+export default class ChannelContentsPage extends PureComponent {
```
dzucconi: So ultimately I don't think we'll do pagination exactly like this: what I'm thinking is that we'll have pages be represented as arrays of IDs and then expose a field that accepts those IDs, that way when you add something, the page pages remain constant. Otherwise we'd have to pass offsets (in addition to page/per params) to the pages which would still work but might me more complex ultimately.

react/components/ChannelContents/components/ChannelContentsPage/index.js
```diff
@@ -0,0 +1,102 @@
...
+        {mode === 'resting' &&
+          <Waypoint onEnter={this.handleEnter} />
+        }
+
+        {mode !== 'active' &&
+          <ChannelContentsPageSkeleton
+            skeleton={skeleton}
+            mode="pending"
+          />
+        }
```
dzucconi: Each page is it's own component with waypoints at its start and end boundary. Once either of those get tripped the page loads and the waypoints are unmounted.

react/components/ChannelContents/components/ChannelContentsPage/index.js
broskoski: Interesting yeah, remembering we already do a basic `channels` query (https://github.com/broskoski/api/commit/5004ba64bbc1595d5e2c186c28653df054a51c7b). I suppose the proposed endpoint would do something similar but in the context of the channel. It would also have to be an array of `type` and `id` objects (to accounts for channel connections).

react/components/ChannelContents/components/ChannelContentsPage/index.js
broskoski: ++

react/components/ChannelContents/components/ChannelContentsPage/index.js
dzucconi: Yep

PR #1967, opened 2019-04-12
Channel progress

dzucconi:
a6a4111 will have effects in production so we'll merge on Monday morning to ensure it's *great success*

### DIFF
--- a/jsconfig.json
+++ b/jsconfig.json
@@ -1,5 +1,10 @@
 {
   "compilerOptions": {
-    "baseUrl": "./"
-  }
+    "baseUrl": "./",
+    "target": "es6"
+  },
+  "include": [
+    "react/**/*"
+  ],
+  "checkJs": false,
 }
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
     "apollo-cache-inmemory": "^1.4.1",
     "apollo-client": "^2.4.10",
     "apollo-link": "^1.2.6",
+    "apollo-link-batch-http": "^1.2.11",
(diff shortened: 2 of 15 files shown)


jsconfig.json
```diff
@@ -1,5 +1,10 @@
...
   "compilerOptions": {
-    "baseUrl": "./"
-  }
+    "baseUrl": "./",
+    "target": "es6"
+  },
+  "include": [
+    "react/**/*"
+  ],
+  "checkJs": false,
```
damassi: @dzucconi - One day y'all will flip on `babel-preset-typescript` and it'll srsly be a happy day for yr future selves.

Even if you adopt the least strict settings, don't add a single typescript-y feature, and write 99% JS, swapping out PropTypes for interfaces will save so much extraneous typing and make things way more DX friendly / safe / simpler. Don't even need to boot the app to do massive cleanups / refactors -- just run `yarn tsc -w` and follow the type-checker. It's very much worth the effort to adopt and the babel preset makes its seamless / instant. 

With the least strict settings you can pretty much rename `.jsx` to `.tsx` and it will just work. Would have to update settings to support absolute paths, and maybe set `libCheck: false`, but not much else. 

If y'all ever end up considering this I'm super down to help lift things over the initial setup hump. Wouldn't take long at all. I've done this so many times that I pretty much have the migration path memorized...

dzucconi: > swapping out PropTypes for interfaces

Yeah, I'm in

damassi: I'm so psyched :) 

Will open a PR with the basics and then we can go from there.

PR #1968, opened 2019-04-12
Browser extension progress

broskoski:
load-extension


extension/src/apollo/index.js
```diff
@@ -0,0 +1,88 @@
...
+    );
+
+export const mountWithApolloProvider = (Component, props = {}, mountNode) => {
+  if (!mountNode) return null;
+
+  const client = initClientSideApolloClient();
+  const WrappedComponent = wrapWithProviders(client)(Component, props);
+
+  return mount(WrappedComponent, mountNode);
+};
```
dzucconi: Why would this need to be duplicated?

extension/src/components/Layout/index.js
```diff
@@ -0,0 +1,97 @@
...
+class Layout extends Component {
+  static propTypes = {
+    children: PropTypes.node.isRequired,
+  }
+
+  constructor(props) {
+    super(props);
+    this.messenger = new Messenger(window.top);
+
+    window.addEventListener('message', this.receiveMessage);
```
dzucconi: In general: add event listeners on mount, remove them on unmount

extension/src/iframe.css
```diff
@@ -0,0 +1,51 @@
...
+  padding: 40px; 
+  -webkit-font-smoothing: antialiased; 
+  text-align: center; 
+  display: inline-block; 
+  font-size: 1.5em;
+}
+
+#arenaExtension_frame ul li.tab--container__nav__item {
+  padding: 0 margin 0 10px !important;
+}
```
dzucconi: Why the CSS? You can export a global style if you want or at the very least have styled components compile the CSS for you for consistency

broskoski: Iframe style is outside the bounds of the react stuff. Trying to keep this to a minimum.

extension/src/apollo/index.js
broskoski: Annoyingly different data sources, most notably no sharify.

extension/src/iframe.css
dzucconi: You could still just pull the stylesheet out a la https://www.styled-components.com/docs/advanced#server-side-rendering

Though if you want to write some artisanal CSS I'd at least use BEM w/ CSS variables (which have fine support)

yarn.lock
```diff
@@ -3661,6 +3666,26 @@ cacache@^10.0.4:
     unique-filename "^1.1.0"
     y18n "^4.0.0"
 
+cacache@^11.3.1:
```
dzucconi: What's this from/for?

extension/src/components/ChannelSelection/components/Search/index.js
```diff
@@ -0,0 +1,23 @@
...
+
+  render() {
+    const { onChange } = this.props;
+
+    return (
+      <SearchInput
+        onQueryChange={onChange}
+        placeholder="Filter channels"
+      />
+    );
```
dzucconi: Something is wrong here the icon isn't being sized correctly; probably because of ems.

dzucconi:
Damon's feedback:

* Clicking "Connect to" takes longer than it should. Very chunky render. It's loading a lot of data that we don't need. I think the 'All channels' section is not necessary and I'm opposed to it. But if we do it we need to really tweak the loading here to be more considerate.
* This view is crazy looking: http://static.damonzucconi.com/_capture/N7Ocv4cXDb.png - We should likely be sizing this thing dynamically to the content. I'm not sure what the restrictions are here? Either way there's a lot we can do to make the link thing look alot better and more consistent with our existing UI
* I actually don't even see the point of the different mode buttons at all (add text, image, etc) since they don't do anything specific; you can do all those things from each of them and they basically look the same. What's the thinking here?

dzucconi:
* Channels should display block and have a healthy clickable area

danpia:
- hovering the list of channels should have a style feedback
- I agree with Damon that dragging images or text is the same and could be a single CTA. Logically you'd land directly on the dragging screen with "Drag images or text here to add to Are.na
" and another button to add page as a link, but I might be missing other design/technical reasons.

dzucconi:
* Filter should autofocus

tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "allowJs": true,
+    "allowJs": false,
```
damassi: This was leading to superr slow type-checking time. No need to process all the js stuff, just stick to typescript code.

src/extension/webpack.config.js
```diff
@@ -62,6 +61,12 @@ const config = {
           },
         ],
       },
+      {
+        // Ensure .mjs files just pass through and are not transpiled
+        test: /\.mjs$/,
+        include: /node_modules/,
+        type: 'javascript/auto',
+      },
```
damassi: /shrug

broskoski:
Gonna merge here and move on!

PR #1973, opened 2019-04-17
Wires up pusher to deal with channel updates

dzucconi:
I'm going to use this packaged version of the Pusher lib in lieu of the CDN one as well. I'll remove that before merging.

pushed

### DIFF
--- a/react/components/ChannelContents/components/ChannelContentsAppended/index.js
+++ b/react/components/ChannelContents/components/ChannelContentsAppended/index.js
@@ -1,0 +1,75 @@
+import React, { useEffect, useState } from 'react';
+import PropTypes from 'prop-types';
+import { withApollo } from 'react-apollo';
+import sharify from 'sharify';
+
+import initPusherClient from 'react/util/initPusherClient';
+
+import channelContentsSetQuery from 'react/components/ChannelContents/components/ChannelContentsSet/queries/channelContentsSet';
+
+import ChannelContentsSet from 'react/components/ChannelContents/components/ChannelContentsSet';
+
+const { data: { NODE_ENV } } = sharify;
+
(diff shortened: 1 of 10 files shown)


broskoski:
🔥

react/components/ChannelContents/index.js
```diff
@@ -54,16 +52,14 @@ export default class ChannelContents extends PureComponent {
...
-                skeleton={newBlocks}
-              />
-            }
           </React.Fragment>
         }
 
+        <ChannelContentsAppended
+          id={id}
+          pendingSkeleton={pendingSkeleton}
+        />
```
dzucconi: This winds up making way more sense: we don't want to rechunk the skeleton and we'd get the exact same result if we did.

broskoski: Nice, yes.

react/components/ChannelContents/components/ChannelContentsAppended/index.js
```diff
@@ -0,0 +1,75 @@
...
+import { withApollo } from 'react-apollo';
+import sharify from 'sharify';
+
+import initPusherClient from 'react/util/initPusherClient';
+
+import channelContentsSetQuery from 'react/components/ChannelContents/components/ChannelContentsSet/queries/channelContentsSet';
+
+import ChannelContentsSet from 'react/components/ChannelContents/components/ChannelContentsSet';
+
+const { data: { NODE_ENV } } = sharify;
```
dzucconi: If testing this locally you have to be mindful of the environment you're running in vs the server's.

dzucconi: By mindful I mean you won't be subscribed to the correct channels.

PR #1976, opened 2019-04-20
Upgrade to Babel 7, Storybooks 5

damassi:
Made a few changes: 

- Upgrades Ervell to Babel 7
- Upgrades Storybooks to 5
- Fixes weird node-gyp error
- Scopes webpack code to `/src/` vs excluding `/node_modules/`

cc @dzucconi


package.json
```diff
@@ -161,14 +170,13 @@
     "faker": "^4.1.0",
     "friendly-errors-webpack-plugin": "^1.7.0",
     "graphql-tools": "^3.0.2",
-    "html5": "*",
```
damassi: This was leading to `node-gyp` errors -- i looked around the codebase and couldn't find anything related, so just removed.

dzucconi: Ah, I tried to track that down a while ago, removing a lot of dependencies and never caught it. Thanks.

damassi: The magic incantation is `yarn why <package-name>`

PR #1977, opened 2019-04-20
Add TypeScript support 

damassi:
This PR adds TypeScript support, in the form of `babel-preset-typescript` as well as some associated tooling. There's a number of changes here:

1) Added `babel-preset-typescript` and configured webpack / node to support `.ts` files.
1) Renamed `src/react` to `src/v2`. By pushing things behind a faux "react" namespace the type-checker was getting confused -- meaning, the npm module `react` doesn't have all of those additional files in the package. This can be renamed to something else if need be, but v2 seemed sufficient to unblock. 
1) Added [lint-staged](https://github.com/okonet/lint-staged) for running linting / typechecking before commit / push. This will ensure that type errors aren't checked into the repo and things are always in a good state. If for whatever reason you want to skip this check, can run the git command with `--no-verify`.
1) Updated `eslint` considerably: 
   - Added [eslint-typescript](https://eslint.org/blog/2019/01/future-typescript-eslint)
   - Dropped support for `airbnb` in favor for `recommended` options. Will hardly notice a difference
   - Added `react/recommended` to the presets and turned off some of the unnecessary ones now that typescript is in.
1) Added [prettier](https://prettier.io/) with default settings. I was very surprised to see this _wasn't_ in Ervell. In 2019 not only is it exceedingly rare to work on a project that doesn't have Prettier installed, but it's also a huge energy waster since thinking about formatting, and performing formatting by hand, makes everything less efficient. (On that note, in some future PR can remove semi-colons if you want since prettier will intelligently insert them in areas where they're required; this setting can be adjusted in `package.json`.) 
1) Improved boot performance. With things inclusively pointing to `/src/`, post-cache warming compiles take about 6s on my machine, and i've noticed that each sub app mounts faster since there are less files in webpack's dev server. 
1) Updated `yarn start-dev` so that terminal formatting can be preserved for cleaner feedback cycles. Right now running everything behind forman clobbers things, and during local dev foreman isn't needed.

### New Workflow

1) `yarn start-dev`
1) This boots Webpack, which runs ForkTSCheckerWebpackPlugin. As changes are made to the codebase this will run in parallel, checking the code
1) When running StoryBooks the same thing happens
1) If performing a refactor, and don't want to boot the app, there's a new `yarn type-check -w` which starts up TypeScripts type-checker

### Overview

- Here's a typical React component, now that hooks are out: 

```tsx
interface FooProps {
  name: string
  optionalDescription?: string
  blockType: "foo" | "bar" | "baz" 
}

export const Foo: React.FC<FooProps> = ({ name, optionalDescription = 'umm', ... }) => {
  return ...
}
```

Most components just need an interface and all of the other props are inferred. Once this is merged I can follow up with another PR that converts a lot of files to `.tsx` and swaps prop-types for interfaces. 

- When wrapping with the Apollo `graphql` HOC:

```tsx
interface FooProps {
  name: string 
}

const Foo: React.FC<FooProps> = ({ name }) => {
  return ...
}

export default graphql<FooProps>(someMutation, {
  name: 'createBlock'
})(Foo)
```

Note that the `graphql` function needs a generic in the form of the component's props, otherwise it doesn't know what's up with it. With HOCs this is often the case. 

(`cmd+hover` over things will show the type signature. That's how I figured out this was needed.)

- **Don't be afraid of `any` as yr getting a hang of TypeScript**. Little things pop up here and there and it's important to think of this as incremental, and will get much better with it in time. 

Using the above example, to skip type-checking on the `graphql` wrapper completely this form could have been used:

```tsx
export default graphql(someMutation, {
  name: 'createBlock'
})(Foo as any)
```

`any` is used to keep momentum going. My workflow both when first getting started with TS up to now is typically:

1) Write everything without types
1) If errors pop up, add `any` and a `// FIXME`
1) Once I'm satisfied with a design, go back and add types. 
1) If I can't figure out a type, I leave the `any` and move on; nbd. 

**One little in general comment** 

Noticed that throughout the code that default exports are used. This is a pattern that should be avoided as default exports aren't statically analyzable as are named exports. 

For example: 

```js
// file A
export const Foo = () => { ... }

// file B
import { Bar } from './Foo' // throws error -- Bar does not exist
```

Additionally, named exports make refactoring stuff way more sane since one can be certain that name / variable mappings always exist 1:1. 

On the subject of imports, TypeScript makes dealing with them really nice 😄 

autoimport


webpack.config.js
```diff
@@ -129,7 +135,21 @@ if (ANALYZE_BUNDLE) {
...
+      checkSyntacticErrors: true,
+      watch: ['./src'],
+    }),
+    new ForkTsCheckerNotifierWebpackPlugin({
+      excludeWarnings: true,
+      skipFirstNotification: true,
+    }),
+    new ProgressBarPlugin(),
+    new WebpackNotifierPlugin()
+  );
```
damassi: Only a few changes were required to webpack file.

damassi:
Alright @dzucconi - this is ready to go (sorry for the large number of changes surrounding the folder rename.)

dzucconi:
Truly great. I don't know what we did to deserve it but I'm very grateful.

Thank you thank you thank you

dzucconi:
This does crash in staging however
```
Error: Cannot find module '/app/src/index.ts'
at Function.Module._resolveFilename (internal/modules/cjs/loader.js:668:15)
at Function.Module._load (internal/modules/cjs/loader.js:591:27)
at Function.Module.runMain (internal/modules/cjs/loader.js:877:12)
at internal/main/run_main_module.js:21:11
```

PR #1978, opened 2019-04-21
Fix index.js boot path

damassi:
Fixes boot location in Procfile. Was pointing at a `.ts` file to see how things would go server-side but then ended up renaming it back so as to do this in another path. 

cc @dzucconi


dzucconi:
👍 👍

PR #1979, opened 2019-04-21
Implement Apollo GraphQL tooling / validation / TypeScript type generation

damassi:
Since y'all are using Apollo there are a few things that can be taken advantage of to make working with graphql a little less free-form, scary and more scalable, but I had to first rename each and every graphql-related fragment / query / mutation file to `.ts` so that type-generation could occur. This was wayy easier thanks to Ervell being so consistent. Once that was in place these changes could happen: 

- Added `yarn generate-types` (and `:watch`) for generating TypeScript types for graphql code. 
- A [vscode-plugin](https://marketplace.visualstudio.com/items?itemName=apollographql.vscode-apollo) for click-to-definition, autocomplete, validation and other goodies. When VSCode boots it will pull down a copy of the remote schema so things are always in sync. Give it a sec to warm up, though.  Requires that a `GRAPHQL_XAPP_TOKEN` is added to ones `.env` file.
- Added `yarn sync-schema` which pulls down the graphql schema from the server and saves it to `/v2/apollo`, if need be.

This is all handled by an `apollo.config.js` file in the root, which interfaces with [apollo-tooling](https://github.com/apollographql/apollo-tooling).

There is one validation error, which I've commented out for now (will note in PR) to get type-generation to work:
- Validation error: `Fragment "KonnectableCell" cannot be spread here as objects of type "User" can never be of type "Konnectable".` This is located in [groupProfileFollowers](https://github.com/aredotna/ervell/blob/master/src/v2/components/ProfileFollows/fragments/groupProfileFollowers.js#L12). @dzucconi - any clue? This seems like a legit schema problem. 
- Note that Apollo doesn't like anonymous graphql queries -- they throw the validator off, so i've given names to each where missing. 

Example of the apollo-tooling in action: 

cmd-hover

schema-validation

### Setup steps:
- Add `GRAPHQL_XAPP_TOKEN` to env file 
- Install the [vscode-plugin](https://marketplace.visualstudio.com/items?itemName=apollographql.vscode-apollo) and restart VSCode. 
- When new `@client` local state additions are made, `v2/apollo/localState/clientSchema.graphql` needs to be updated.


damassi:
This is getting close. It looks like the remaining errors involve client local schema directives `@client`.

apollo.config.js
```diff
@@ -0,0 +1,23 @@
+require('dotenv').config();
+
+// Configuration for the vscode-apollo extension.
+module.exports = {
+  client: {
+    service: {
+      name: 'are.na',
+      url: 'https://api.are.na/graphql',
+      headers: {
+        'X-APP-TOKEN': process.env.GRAPHQL_XAPP_TOKEN,
```
damassi: This is where apollo's tooling pulls in the schema. I suspect the issue with `@client` is that it's pulling a remote schema which has no notion of `@client` and then failing to recognize client-only additions. I'll need to play with it a bit more.

typings/global.d.ts
```diff
@@ -0,0 +1,18 @@
...
+
+declare module '*.svg' {
+  const content: any;
+  export default content;
+}
+
+declare global {
+  interface Window {
+    __APOLLO_CLIENT__: any;
+    __APOLLO_STATE__: any;
```
damassi: When needing to add stuff to globals like window, etc, it happens in this file.

apollo.config.js
```diff
@@ -0,0 +1,16 @@
...
+      },
+    },
+    includes: ['src/**/*.ts', 'src/v2/apollo/localState/clientSchema.graphql'],
```
damassi: This pulls in the `clientSchema` so that we can generate types and get validation and so on. Note that I typed each property in here as a string for now just to get things working, but as code is converted to TypeScript these types will likely need to be updated to match how they're actually used in code.

package.json
```diff
@@ -7,13 +7,16 @@
     "deploy-staging": "sh scripts/deploy-staging.sh",
     "deploy-storybook": "storybook-to-ghpages",
     "deploy": "sh scripts/deploy.sh",
+    "generate-types": "apollo codegen:generate --target typescript --outputFlat src/v2/apollo/types",
+    "generate-types:watch": "yarn generate-types --watch",
```
damassi: Once more TypeScript is added to / converted in the codebase and graphql queries are being written on the fly, a future workflow item (to avoid starting processes in two separate tabs) would be to add this to package.json scripts:

```
"start-dev": "concurrently --raw --kill-others 'yarn start' 'yarn generate-types:watch'"
```
The server boots, and then apollo codegen is in the background spinning generating types as new stuff comes along or as graphql code changes. VSCode would then pick up the changes.

package.json
```diff
@@ -7,13 +7,16 @@
...
     "jest": "jest",
     "lint": "eslint --ext .js,.jsx,.ts,.tsx ./src --fix",
     "mocha": "sh scripts/mocha.sh",
     "predeploy": "sh scripts/predeploy.sh",
     "restart-dyno": "node bin/restart",
     "start-dev": "node src/index.js",
     "storybook": "start-storybook -p 9001 -c .storybook",
+    "sync-schema": "babel-node scripts/syncSchema.js",
```
damassi: Noticed y'all have a `schema.graphql` and a `schema.json` file in `v2/apollo`. Figured this would be a handy script to pull updates in.

package.json
```diff
@@ -40,6 +43,9 @@
       "git add"
     ]
   },
+  "resolutions": {
+    "graphql": "^14.1.1"
+  },
```
damassi: Apollo needs to be sure there's exactly one instance of `graphql` being used in the tree, so that version mismatches don't occur.

package.json
```diff
@@ -7,13 +7,16 @@
...
+    "generate-types": "apollo codegen:generate --target typescript --outputFlat src/v2/apollo/types",
+    "generate-types:watch": "yarn generate-types --watch",
     "jest": "jest",
     "lint": "eslint --ext .js,.jsx,.ts,.tsx ./src --fix",
     "mocha": "sh scripts/mocha.sh",
     "predeploy": "sh scripts/predeploy.sh",
     "restart-dyno": "node bin/restart",
     "start-dev": "node src/index.js",
     "storybook": "start-storybook -p 9001 -c .storybook",
+    "sync-schema": "yarn run apollo schema:download src/v2/apollo/schema.json",
```
damassi: Saw that y'all have a `schema.json` and `schema.graphql` in the `v2/apollo` folder. Figured this would be a handy script to have. Hooks into `apollo.config.js` in the project root, and requires that a `GRAPHQL_XAPP_TOKEN` be set in `.env`.

package.json
```diff
@@ -7,13 +7,16 @@
     "deploy-staging": "sh scripts/deploy-staging.sh",
     "deploy-storybook": "storybook-to-ghpages",
     "deploy": "sh scripts/deploy.sh",
+    "generate-types": "apollo codegen:generate --target typescript --outputFlat src/__generated__",
```
damassi: GraphQL types are output to an `src/__generated__` folder.

src/v2/components/ProfileFollows/fragments/groupProfileFollowers.ts
```diff
@@ -8,8 +8,8 @@ export default gql`
     __typename
     id
     followers(page: $page, per: $per) {
-      ... IdentifiableCell
-      ... KonnectableCell
+      ...IdentifiableCell
+      # ...KonnectableCell
```
damassi: @dzucconi - in order to get type generation to work I had to comment out this validation error:

```
Fragment "KonnectableCell" cannot be spread here as objects of type "User" can never be of type "Konnectable".
```

I'm pretty sure this will break something somewhere, but did seem like a legit error. Thoughts? 

This also occurs in `src/v2/components/ProfileFollows/fragments/userProfileFollowers.ts`.

src/v2/apollo/index.tsx
```diff
@@ -91,6 +92,36 @@ export const initApolloClient = ({
...
+    }
+
+    extend type ClientSerializedMe {
+      __typename: String
+    }
+
+    extend type ClientSharify {
+      __typename: String
+    }
+  `;
```
damassi: This bit was been removed in a later commit.

.vscode/settings.json
```diff
@@ -5,15 +5,15 @@
     "public/": false
   },
   "search.exclude": {
+    "**/__generated__": false,
```
damassi: I thought about excluding stuff in this folder from fuzzy match / search, but realized that that would then limit the ability to do TypeScript auto-imports, e.g.,: 

Screen Shot 2019-04-21 at 10 56 52 PM

damassi:
Alright this work is now complete. TypeScript type generation is fixed, a ton of sensitive files have been renamed to to .ts (all of y'alls graphql query / fragment / mutations -- which makes things quite a bit safer overall since it's the data layer) and a few other files have been converted to `.tsx`. Here's a rundown: 

- The main issue, which was a serious journey to figure out, was figuring out how to get type generation and VSCode tooling to recognize `@client` directives. This was solved in [this commit](https://github.com/aredotna/ervell/pull/1979/commits/6c6d29264e294a12e610d49f627f1a4bfdf28878#diff-9e94de47c5f91771fb0e8d84cf58171bR1). Note that as additional client local state is added, this file will need to be updated accordingly. 

- This will now allow us to remove all of those `graphql-anywhere` propTypes functions which infer data based on GraphQL fragments, in favor or proper interfaces. Much nicer 😄.  

- Generated GraphQL types were added to `src/__generated__` in [this commit](https://github.com/aredotna/ervell/pull/1979/commits/008977b38d32614173516197d99c4d4a112669e9). Run `yarn generate-types` to update, and `yarn generate-types:watch` while actively developing. 

- An example of how to use these types can be seen in [this commit](https://github.com/aredotna/ervell/pull/1979/commits/9a80503f02c02343626e9ae235e974f1cefe634a#diff-8f06ee2099b1edb70b9466b16cf7911fR9). This will be the way to do it going foward, and was the whole point of this PR / getting TypeScript in -- so all of this data can now be TypeSafe and maintainable over time. 

- One thing to keep in mind is oftentimes the component name is the same as the file name in `__generated__` so if you start typing expecting to see the auto-import popup, it won't be there, since there's already that symbol in scope. In the bullet above, I temporarily changed the name of the component to get the auto-import, and then use `as Something` to assign it a unique import name. A lot of the time i'm sure this will be a non-issue tho. 

- Apollo VSCode shows that y'all are using a couple deprecations; might want to take care of that to silence the warnings: 

Screen Shot 2019-04-21 at 11 03 29 PM

-**Lastly, the issue surrounding `Konnectable` will need to be fixed** either in this PR or in a follow-up. I'm not familiar enough with your schema to address it so I've simply commented it out. Are.na still boots tho /shrug. Here's the error again: `Fragment "KonnectableCell" cannot be spread here as objects of type "User" can never be of type "Konnectable".`

.vscode/settings.json
damassi: (Also note the `SpaceProps` in there -- from `@types/styled-system`. Every styled-system function has a interface like this that you can import. I would suggest going into yr UI folder and typing all of your primitives following a pattern similar to Palette's [`<Flex>` component](https://github.com/artsy/palette/blob/master/packages/palette/src/elements/Flex/Flex.tsx#L46-L90)

dzucconi:
@damassi just getting around to booting this today. What's your total mount time on your machine? I'm thinking we're going to have to precompile `src` just to get deploys up without noticeable downtime.

damassi:
It's still pretty lengthy, all said.... 

```
actions: 966.367ms
feed: 13436.390ms
home: 230.235ms
blog: 13.058ms
authentication: 1007.235ms
confirmation: 9.381ms
getting_started: 3.474ms
tools: 37.085ms
util: 6.067ms
about: 762.794ms
search: 463.686ms
manage: 6.989ms
share: 69.248ms
marklet: 45.618ms
import: 124.061ms
settings: 50.906ms
onboarding: 540.372ms
graphql: 29.344ms
examples: 22.639ms
profile: 3264.299ms
block: 186.413ms
new_channel: 3392.223ms
channel: 232.456ms
rss: 16.299ms
```

That should def take place in a different PR tho since this one covers so much. I could pretty easily port over work I did [in force around this](https://github.com/artsy/force/pull/3630) later today if that would help.

dzucconi:
Yah, the compilation speed isn't a deal breaker because it was already slow to begin with, but it did get quite a bit *slower*

damassi:
Oh I didn't realize it was slower, 👀 at some previous commits to compare. (Might just be babel's cache warming up during first compile, though a cache wouldn't help if y'alls deploys are containerized over at Heroku or wherever.)

damassi:
I just jumped back to `8d1c9875f9c373229e9c4df247068deae02da788` (just prior to starting this recent batch of work) and booted up the server and here are the numbers: 

```
about: 661.629ms
actions: 1023.165ms
authentication: 1030.632ms
block: 148.099ms
blog: 15.887ms
channel: 277.461ms
confirmation: 18.026ms
examples: 46.023ms
feed: 11745.044ms
getting_started: 6.447ms
graphql: 23.073ms
home: 367.249ms
import: 52.673ms
manage: 13.731ms
marklet: 32.117ms
new_channel: 2543.571ms
onboarding: 538.527ms
profile: 2670.495ms
rss: 27.881ms
search: 531.364ms
settings: 36.412ms
share: 77.417ms
tools: 48.294ms
util: 7.354ms
```

It's all very ball-park, with some being very slightly more and some being very slightly less. If everything was averaged out it would likely be the same or only marginally faster / slower in each direction. The perception that it's a lot slower might be a perceptual trick having to do cleaning  up the default webpack [output in here](https://github.com/damassi/ervell/blob/4f678f8fdeae3d8ea0d44c3a2079c9f942c18a5f/src/lib/webpack-dev-server.js#L12). 

From a client-side compilation perspective pre-cache comilation is about 10s faster on my machine, due to babel 7 -- around 23-26s, and then  once it's warmed up its between 5s and 7s. That might also contribute since it throws off some of the time-scales y'all have gotten used to working with.

damassi:
(rebased from master)

dzucconi:
Just re those compile times: I don't think you're counting the `webpack built` time that happens before  anything is even mounted. That's the time that I'm seeing a dramatic increase in.

dzucconi:
Part of my concern with this at the moment is that the hot reloading breaks all the time in the app dev process. If I'm working on integration (not in Storybooks), killing and restarting the server happens 
 frequently enough for this to become a pretty big problem. I'm wondering if there's a way to see where a more granular breakdown of what's eating time/resources during compilation?

damassi:
@dzucconi - So the issues seen are

- [ ] Hot reloading is inconsistent. This might be related to me removing the hot-reload code from the babel-file in a previous PR. I noticed in force that `hot-reloader` and so on is not longer needed with latest webpack, but let me double check. This is fixable, will take a look now.
- [ ] The error that's thrown with IconicJS lib, when clicking into a channel. Will fix that now as well. 

Initial compile time is quite a bit faster on my machine with babel 7. The general flow is: when running `yarn install` -- after deleting node_modules completely -- the babel cache is cold. Running `yarn start-dev` then compiles, creating a new cache. Subsequent builds (assuming changes haven't been made to critical code paths requiring an invalidation) are then in the 5-7s range, which is very good for an app of this size. This isn't consistent with what you're seeing? 

Will give the two issues above a look now.

damassi:
Here's what i'm seeing reload-time-wise which is nearly instant: 

reload

What file are you specifically editing? I've checked a few places now and every place reloads the same. Maybe run `yarn --check-files`?

damassi:
Fixed the Iconic issue. using `window.IconicJS` fixes the problem. Not sure why this was needed, but def some weird old vendored js issue. Since this is old code wouldn't worry about it too much.

src/apps/about/index.js
```diff
@@ -13,19 +13,18 @@ const app = express();
...
-  })
-  .catch(next);
+const resolveRoadmap = (req, res, next) =>
+  client
+    .getEntry(ROADMAP_ENTRY_ID)
+    .then(entry => req.apollo.render(RoadmapPage, { roadmap: entry }))
+    .then(apollo => {
+      res.render('roadmap', { apollo });
+    })
+    .catch(next);
```
damassi: just prettier changes from touching the file after adding a log in

PR #1982, opened 2019-04-22
Add regenerator runtime

damassi:
Babel 7 handles polyfill stuff differently than babel 6, via the `preset-env` plugin. See https://babeljs.io/docs/en/babel-preset-env#usebuiltins. 

That said, sometimes the regenerator will still need to be imported. I can take a better look at this later but this should get things to a deployable state.


dzucconi:
🙇

PR #1984, opened 2019-04-23
Minor error handling fix; changes x-app token env var

dzucconi:


dzucconi:
Hm, prettier is nice but maybe we should run it over the whole project to limit future diff carnage?

PR #1986, opened 2019-04-23
Precompile server code

damassi:
Adds a new webpack config (`serverConfig.js`) which precompiles server-side code into a flat `server.dist.js` file. This fixes boot times since individual apps no longer need to be compiled on the fly. 

New command:
```
yarn build:server
```

This can be optimized in the future, but to test a local prod build:
- Run `yarn build:server`
- Run `yarn bucket-assets --fingerprint false`
- Copy the output from `manifest.json` and paste into `ASSET_MANIFEST` in .env file
- Add `CDN_URL=''` to `.env`
- Run `NODE_ENV=production yarn start:prod`

boot

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
   "scripts": {
     "assets": "sh scripts/assets.sh",
     "build": "yarn webpack --config ./webpack/webpack.config.js",
+    "build:server": "BUILD_SERVER=true NODE_ENV=production yarn build",
     "deploy-staging": "sh scripts/deploy-staging.sh",
     "deploy-storybook": "storybook-to-ghpages",
     "deploy": "sh scripts/deploy.sh",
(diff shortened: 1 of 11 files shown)


dzucconi:
Ah alright, so it's the bucketAssets middleware that's causing this to hang:
https://github.com/aredotna/bucket-assets/blob/d253febf7c658673c036afc85d034572011135e4/index.js#L19

package.json
```diff
@@ -16,6 +17,7 @@
     "predeploy": "sh scripts/predeploy.sh",
     "restart-dyno": "node bin/restart",
     "start": "node --max_old_space_size=4096 -r dotenv/config src/index.js",
+    "start:prod": "node --max_old_space_size=4096 -r dotenv/config server.dist.js",
```
damassi: I added this for testing, but might want to clean up all of these `start` scripts (and below). `start-dev` and `start-prod` are just pointers to other env files as an example of how its done.

package.json
```diff
@@ -15,7 +16,8 @@
     "mocha": "sh scripts/mocha.sh",
     "predeploy": "sh scripts/predeploy.sh",
     "restart-dyno": "node bin/restart",
-    "start": "node --max_old_space_size=4096 -r dotenv/config src/index.js",
+    "start": "node -r dotenv/config src/index.js",
```
damassi: I added this for testing, but might want to clean up all of these `start` scripts (and below). `start-dev` and `start-prod` are just pointers to other env files as an example of how its done.

package.json
```diff
@@ -15,7 +16,8 @@
...
+    "start:prod": "node -r dotenv/config server.dist.js",
```
damassi: I added this for testing, but might want to clean up all of these `start` scripts (and below). `start-dev` and `start-prod` are just pointers to other env files as an example of how its done.

PR #1990, opened 2019-04-26
Runs prettier on src/v2/**/*

dzucconi:
I just really wanted to write all this code


dzucconi:
OK https://github.com/aredotna/ervell/pull/1991 wins

PR #1991, opened 2019-04-26
Run prettier over project

damassi:
Runs a quick Prettier sweep so that PR diffs don't need to be muddied up. Also adds a `.prettierignore` which at the moment ignores generated type files in `src/__generated__`.


.prettierignore
```diff
@@ -0,0 +1 @@
+src/__generated__/**
```
dzucconi: A ha

PR #1992, opened 2019-04-26
Fixes webpack dev config

dzucconi:
Re: https://www.styled-components.com/docs/faqs#why-am-i-getting-a-warning-about-several-instances-of-module-on-the-page

### DIFF
--- a/webpack/envs/baseConfig.js
+++ b/webpack/envs/baseConfig.js
@@ -85,6 +85,8 @@ const baseConfig = {
     alias: {
       'jquery.ui.widget': 'blueimp-file-upload/js/vendor/jquery.ui.widget.js',
       Images: path.join(rootDir, 'public', 'images'),
+      'styled-components': require.resolve('styled-components'),
+      react: require.resolve('react'),
     },
     extensions: [
       '.js',


webpack/envs/baseConfig.js
```diff
@@ -102,6 +102,21 @@ const baseConfig = {
...
+      cacheGroups: {
+        runtimeChunk: 'single',
+        vendor: {
+          name: 'vendor',
+          chunks: 'initial',
+          test: /[\\/]node_modules[\\/]/,
+        },
+      },
+    },
+  },
```
damassi: Before making this change, try updating the base webpack config with these two settings, under `resolve.alias`: 

https://github.com/artsy/force/blob/master/webpack/envs/baseConfig.js#L81-L82

This ensures that things resolve to a single instance. The react one is also important because a similar problem happens with hooks. (The _underlying_ problem is that with React 16.3's new context api, context usage in React and Styled Components is now static, derived from a top level export, rather than an instance, derived from `new`.)

damassi: Not relevant to this PR but good to know if you ever plan on breaking out the UI system into it's own package and want to use `yarn link` in combination with SSR. This is what fixed it for us: 

https://github.com/damassi/force/blob/7b13d63872d4d5748029def19f5dcd9ae25f0841/src/index.js#L31-L42

dzucconi: Much thanks!

PR #1993, opened 2019-04-26
Run `yarn type-check` on CI

damassi:
Adds a `type-check` step the CI so that errors don't slip through in commits that may have been run with `--no-verify`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,15 @@
 language: node_js
 node_js:
-- '8'
+  - '8'
 script:
-- yarn jest
-- yarn lint
+  - yarn jest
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,14 +1,15 @@
 language: node_js
 node_js:
-- '8'
+  - '8'
 script:
-- yarn jest
-- yarn lint
+  - yarn jest
+  - yarn lint
+  - yarn type-check
```
damassi: This is the only change; rest is prettier

dzucconi: 👍

PR #1997, opened 2019-04-27
Remove bucket-assets, replace with Webpack hashing

damassi:
This PR removes `bucket-assets` in favor of Webpack and a local S3 upload script.

#### Overview: 

1) Replaced un-cacheable bucket-asset fingerprints with consistent hashes from webpack (hash only changes when the content inside changes, and webpack will track that)
1) This gives us source maps, since the hash from webpack matches the hashed source-map filename, also generated by webpack (rather than an external lib)
1) Replaced bucket-assets upload functionality with a local script, `scripts/uploadToS3.js`
1) Added a script to compile / hash CSS, which will change per build (like it did before). This doesn't go through Webpack because the code is too old and it would be hell to fix; it's mostly irrelevant as the app is now largely styled-components. The hashes get merged into webpacks manifest as a seed
1) Added a new middleware helper to return manifest assets
1) `ASSET_MANIFEST` env var can be removed from local envs and Heroku

To run / boot a full prod build: 
```
NODE_ENV=production yarn build && yarn build:server && yarn upload-to-s3 && yarn start:prod
```

(If needed, append `dotenv_config_path=./.env.production` to the script up above, but make sure that `NODE_ENV=production` is set in `.env` file.) 

To test assets running locally, comment out `CDN_URL` in `.env` file.

### DIFF
--- a/scripts/predeploy.sh
+++ b/scripts/predeploy.sh
@@ -3,7 +3,4 @@
 set -e -x
 
 yarn assets
-gzip -S .cgz $(find public/assets -name '*.css')
-gzip -S .jgz $(find public/assets -name '*.js')
-bucket-assets --bucket $S3_BUCKET
-heroku config:set ASSET_MANIFEST=$(cat manifest.json) --app=$APP_NAME
+yarn upload-to-s3
--- a/webpack/utils/getCSSManifest.js
+++ b/webpack/utils/getCSSManifest.js
@@ -1,0 +1,79 @@
+/* eslint-disable no-console */
+
+const chalk = require('chalk')
+
+// @ts-check
+
+const glob = require('glob')
+const fs = require('fs')
+const path = require('path')
+const crypto = require('crypto')
+const { execSync } = require('child_process')
+const baseConfig = require('../envs/baseConfig')
+
+// Ouput
+const DEST = 'public/assets'
+
+// Tasks
+function clean() {
+  glob.sync(`${DEST}/*.{css,css.cgz}`).forEach(file => fs.unlinkSync(file))
+}
+
+function compile() {
+  console.log(chalk.green(`[Ervell compileCSS] Compiling...`))
+  const files = glob.sync('src/assets/*.styl', {
+    nodir: true,
+  })
+  execSync(`yarn stylus ${files.join(' ')} --compress -o ${DEST}`)
+}
+
+function fingerprint(file) {
+  const contents = fs.readFileSync(file)
+  const hash = crypto
+    .createHash('sha1')
+    .update(contents)
+    .digest('hex')
+    .slice(0, 21)
+  const ext = '.css'
+  const original = file.replace('public', '')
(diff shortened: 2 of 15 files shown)


scripts/predeploy.sh
```diff
@@ -3,7 +3,4 @@
 set -e -x
 
 yarn assets
-gzip -S .cgz $(find public/assets -name '*.css')
-gzip -S .jgz $(find public/assets -name '*.js')
-bucket-assets --bucket $S3_BUCKET
-heroku config:set ASSET_MANIFEST=$(cat manifest.json) --app=$APP_NAME
+yarn upload-to-s3
```
damassi: There's no need to gzip locally -- all of this can be taken care of in Cloudfront settings:

1) find your distro, and click the Behaviors tab

Screen Shot 2019-04-26 at 8 54 01 PM

2) Hit edit
3) Flip on the automatic compression setting: 

Screen Shot 2019-04-26 at 8 53 49 PM

damassi:
Since I don't have access to ya'lls AWS S3 bucket I can't test this end-to-end, but I _think_ this is good to go as is. We can work out any issues if they come up.

webpack/utils/getCSSManifest.js
```diff
@@ -0,0 +1,79 @@
...
+    clean()
+    compile()
+    const manifest = createManifest()
+    console.log(chalk.green(`[Ervell compileCSS] Complete.`))
+    return manifest
+  } catch (error) {
+    console.error(chalk.red('[Ervell compileCSS] Error:', error))
+    process.exit(1)
+  }
+}
```
damassi: We still have lots of CSS in force so this was necessary -- not sure how relevant it is in Ervell since there's only two .styl files.

scripts/predeploy.sh
dzucconi: 👍

dzucconi:
This looks good from here; will move to staging

dzucconi:
Deep thanks for porting this work over. 🎈 🥇

damassi:
Hope it works!

dzucconi:
@damassi works great. Up in staging now.

Sidenote: I've definitely lost the plot on how our static assets get managed along the way. @broskoski do you have any idea what is setting `IMAGE_PATH`? It looks like it's keyed to dd75f5c which is in 2015??? Presumably these aren't going to change anytime soon but it'd be nice to have a plan here.

damassi:
Be sure to flip on that CloudFront setting before shipping to prod: 

Screen Shot 2019-04-29 at 9 36 09 AM

(See the far right column --no gzip.)

dzucconi:
@damassi yeah I turned it on but I think I had to also whitelist the accepts-encoding header; it should be updating now.

damassi:
Might need to send out an invalidation for `/assets/*` before it updates.

dzucconi:
All good now

PR #1998, opened 2019-04-28
Fix outbound links in Lightbox description

damassi:
Noticed that description links in the new lightbox don't click out anymore and instead trigger the edit modal, which is slightly less useful. This checks to to see if the underlying element is an `A` and then catches it, deferring to the outbound description link behavior.

#### Before: 

before

#### After: 

after

### DIFF
--- a/src/v2/components/BlockLightbox/components/BlockLightboxMetadataPane/index.js
+++ b/src/v2/components/BlockLightbox/components/BlockLightboxMetadataPane/index.js
@@ -109,9 +109,16 @@ export default class BlockLightboxMetadataPane extends PureComponent {
             breakWord
             boldLinks
             hoverLinks={{ color: 'black' }}
-            onClick={
-              block.can.manage ? this.openManageFor('description') : undefined
-            }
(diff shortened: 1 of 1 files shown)


src/v2/components/BlockLightbox/components/BlockLightboxMetadataPane/index.js
```diff
@@ -109,9 +109,16 @@ export default class BlockLightboxMetadataPane extends PureComponent {
...
-              block.can.manage ? this.openManageFor('description') : undefined
-            }
+            onClick={event => {
+              // If clicking an A tag in the description don't open edit box and
+              // instead defer to outbound link.
+              if (event.target.nodeName === 'A') {
+                return
+              }
+              if (block.can.manage) {
+                return this.openManageFor('description')(event)
```
damassi: This check could go in `openManageFor` but this behavior seemed fairly specific to the description area; can update if need be.

dzucconi:
Thanks for this!

PR #2000, opened 2019-04-29
Fixes Promise in RSS feeds

dzucconi:
YandexBot is causing all these routes to throw

`TypeError: Promise.all(...).then(...).catch(...).done is not a function`

----

These should
- [ ] Be uncrawlable in the robots.txt
- [ ] Rewritten to fetch using GraphQL
- [ ] Rewritten to use TypeScript


dzucconi:
As far as I can tell this is due to `global.Promise = require('bluebird')` not working in CoffeeScript.

PR #2001, opened 2019-04-29
Removes Iconic injection

dzucconi:
The sizing here is breaking and this is going to be rough to debug. I'm removing this in favor of ignoring the issue until we launch new channels. @broskoski feel free to fix if you feel the need.


dzucconi:
I'd rather disable this now than rollback.

PR #2006, opened 2019-05-01
ContextMenu

dzucconi:
contextdemo

I have a bunch of typing to clean up that I punted on and a few minor things to wrap up but this is the bulk of it.

- [ ] Flesh out types
- [ ] Add in confirm for removing connection


.eslintrc.js
```diff
@@ -29,6 +29,8 @@ module.exports = {
     },
   },
   rules: {
+    'react-hooks/rules-of-hooks': 'error',
+    'react-hooks/exhaustive-deps': 'warn',
```
dzucconi: @damassi I've been finding this rule very helpful for useMemo/useCallback if you haven't seen it yet 📦

damassi: Oh yeah, I use it on some personal projects! In artsy ecosystem we're still on TSLint tho -- need to make the switch over to eslint pretty soon, def in the plans.

dzucconi: What is the deal with tslint/eslint anyway?

src/v2/components/Cell/components/Konnectable/index.tsx
```diff
@@ -50,18 +47,18 @@ const Comments = styled(Typography).attrs({
...
-    ),
-    isPreviewable: PropTypes.bool,
-  }
+interface ContextProps {
+  __typename: string
+  id: number
+}
+
+interface Props {
+  konnectable: KonnectableCellInterface
```
damassi: Nice to see this is working well

.eslintrc.js
damassi: The typescript team recently decided to throw their support behind eslint as they try to scale TypeScript to the wider JS ecosystem, vs extending support for TSLint. Mostly for interop with existint eslint rules, but there were also some architectural issues in TSLint that makes it less performant. 

However, running `eslint-typescript` on a handful of staged files feels wayy slower than TSLint, that's why I removed it from `package.json` in [this commit](https://github.com/aredotna/ervell/commit/42b728512b7dea4d3fe0fa33959dfce3897b18e5#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L44). That said, running the linter in a precommit hook catches a ton of issues at scale and over time and I would recommend adding it back if yr cool with adding a few more seconds to your commit times. With all of the other changes I didn't want to slow down y'alls workflow too much, but its worth adding it. Seems pointless since `--fix` is turned on the editor but there's lots of things that `--fix` doesn't catch and having totally clean commits adds up.

But yeah check out this post for a few links and so on; I wrote about this recently: 
https://artsy.github.io/blog/2019/01/29/from-tslint-to-eslint/

damassi: Speaking of TypeScript, Orta is leaving Artsy for a specially-created spot on the TypeScript team. The way I understand it he's gonna be like the Dan Abromov of TS world. Pretty cool.

damassi: Just checked that Travis failure and that's a good example -- having it in the precommit hook would have caught that before it had a chance to fail on CI.

PR #2008, opened 2019-05-05
Update README with new start commands

damassi:
Removes references to unneeded libs and updates yarn start commands.


dzucconi:
Thanks!

PR #2019, opened 2019-05-10
Fixes notification link for comments / sets up jest with TypeScript

broskoski:
Was going to do this all client side, decided to write a test, spent a long time getting jest and enzyme to work with TypeScript, and then realized a better way to solve in the API (here: https://github.com/aredotna/api/pull/1043). 

In any case, tests for TypeScript are setup, will finish and merge this when the API is deployed.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,0 +1,19 @@
+module.exports = {
+  roots: ['<rootDir>/src'],
+  transform: {
+    '^.+\\.jsx?$': 'babel-jest',
+    '^.+\\.tsx?$': 'babel-jest',
+  },
+  testURL: 'http://localhost/',
+  setupTestFrameworkScriptFile: 'jest-enzyme',
+  testEnvironment: 'enzyme',
+  testEnvironmentOptions: {
+    enzymeAdapter: 'react16',
+  },
+  modulePaths: ['/src', '/node_modules'],
+  moduleFileExtensions: ['js', 'jsx', 'ts', 'tsx'],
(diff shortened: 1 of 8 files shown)


jest.config.js
```diff
@@ -0,0 +1,19 @@
+module.exports = {
+  roots: ['<rootDir>/src'],
+  transform: {
+    '^.+\\.jsx?$': 'babel-jest',
+    '^.+\\.tsx?$': 'babel-jest',
+  },
```
damassi:
```js
transform: {
  "^.+\\.(ts|tsx|js|jsx)$": "babel-jest",
}
```

jest.config.js
```diff
@@ -0,0 +1,19 @@
...
+  testURL: 'http://localhost/',
+  setupTestFrameworkScriptFile: 'jest-enzyme',
```
damassi: This is typically a pointer to a physical file where additional setup can take place. See https://github.com/artsy/force/blob/master/test.config.js

package.json
```diff
@@ -234,6 +240,7 @@
     "sinon": "*",
     "source-map-support": "^0.5.12",
     "storybook-react-router": "^1.0.5",
+    "ts-jest": "^24.0.2",
```
damassi: This should be covered by `babel-jest`, in combination with the `@babel/preset-typescript` (though I imagine this is just experiment code)

broskoski:
Screenshot 2019-05-10 13 19 40

Issue #2009, opened 2019-05-06
Channel TODO


dzucconi:
## Remaining

- [x] SSR the first 'page' of connectables
- [x] Channel connectables filter
- [x] Shareable channel view
- [x] ["You've reached your limit"](http://static.damonzucconi.com/_capture/ITGttxlz3r.png) add block CTA
- [x] Rewire Pusher updates
- [x] Followers page
- [x] Metatags (title, etc)
- [x] Hide AddBlock CTA on private share channels by using the new can field
- [x] [Empty channel message](http://static.damonzucconi.com/_capture/yMjTXcd42c.png)
- [x] Embed view (/embed)
- [ ] Confirm connection destruction in context menu

PR #2021, opened 2019-05-10
Rollout new_channel to admins

dzucconi:
It just occurred to me that I should include the inverse of this on new_channel incase any of these links get shared prematurely.


dzucconi:
A couple of minor issues to get through https://github.com/aredotna/ervell/issues/2009

PR #2026, opened 2019-05-15
Adds instant.page

dzucconi:
@damassi any gotchas here I should know about?


damassi:
Nope, it's sturdy. A previous version used to error (but not break) on `a` tags that lacked `hrefs`, but that was fixed a while back. No other issues.

PR #2027, opened 2019-05-16
Extension redux

broskoski:
extension-example

This goes back to the older style, but with block editing functionality. Going to punt on the keyboard cursors for now, as I can't get the iframe to focus the text field automatically....🙃


src/v2/styles/colors.js
```diff
@@ -12,6 +12,7 @@ const COLORS = {
     neutral: '#fdffdb', // alias editable
     investor: 'rgb(183, 158, 110)',
     supporter: '#00a06d',
+    highlighted: '#fdfdd9',
```
dzucconi: This is suuuuuper close to fdffdb, just use that and/or rename

src/extension/src/iframe.css
```diff
@@ -20,30 +20,31 @@
...
+  padding: 1em 0.5em;
+  -webkit-font-smoothing: antialiased;
+  text-align: center;
+  display: inline-block !important;
   font-size: 1.25em;
   border: dashed 2px #ddd !important;
   margin: 0 auto;
-}
\ No newline at end of file
+}
```
dzucconi: I still think that this should be generated using styled-components.

src/extension/src/lib/Pane.js
```diff
@@ -62,11 +66,16 @@ class Pane {
   }
 
   focusIframe = () => {
-    this.frame.contentWindow.focus()
+    // var frame = this.frame
+    // this.frame.contentWindow.focus()
+    // setTimeout(() => {
+    //   frame.contentWindow.focus()
+    // }, 100)
```
dzucconi: Can delete

broskoski:
Just waiting for @danpia to advise me on what to do with the current empty state, and then this should be ready to go.

PR #2033, opened 2019-05-17
Fully typed Cell.Konnectable

dzucconi:


dzucconi:
+ converting to functional components where relevant

PR #2038, opened 2019-05-21
Implements some minor rendering optimizations

dzucconi:
I need to write a real benchmark — I do wish there was some decent off the shelf tooling for this but it looks like a ton of things have been deprecated in favor of Chrome tools that I don't find particularly useful.

### DIFF
--- a/src/v2/components/ChannelContents/index.tsx
+++ b/src/v2/components/ChannelContents/index.tsx
@@ -255,7 +255,6 @@ const ChannelContents: React.FC<ChannelContentsProps> = ({
                   <ChannelContentsItem
                     key={connectableKey}
                     index={connectableIndex + pageIndex * chunkSize}
-                    connectableSkeleton={connectableSkeleton}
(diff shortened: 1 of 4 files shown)


src/v2/components/ChannelContents/index.tsx
```diff
@@ -41,244 +47,243 @@ interface ChannelContentsProps extends Props {
...
+    // Used to load/unload waypoints
+    const [activeQueries, setActiveQueries] = useState<
+      ActiveQueriesCollection.ActiveQueries
+    >({})
+
+    // Handles ordering of block grid items
+    const skeleton: ChannelContentsSkeletonData[] = isClientSide
+      ? channel.skeleton
+      : first(channel.skeleton, chunkSize)
+    const [connectables, setConnectables] = useState(skeleton)
```
dzucconi: This appears to work pretty seamlessly locally. Will see if it has any real effect 🤞

PR #2039, opened 2019-05-21
Change log_out to POST req

broskoski:


src/v2/components/UserDropdown/components/SecondaryLinks/components/SignOutLink/index.js
```diff
@@ -1,11 +1,17 @@
...
 
 const signOut = () => {
   window.localStorage.clear()
-  window.location.href = '/me/sign_out'
+  axios({
+    method: 'POST',
+    url: `/me/sign_out`,
+  }).then(() => {
+    window.location.href = ''
+  })
```
dzucconi: That'll do it

PR #2041, opened 2019-05-21
Cancel model opening if text is selected

broskoski:
block-text-selection

Fixing an issue reported in Slack
Screen Shot 2019-05-21 at 6 04 26 PM

### DIFF
--- a/src/v2/components/BlockLightbox/components/BlockLightboxText/index.js
+++ b/src/v2/components/BlockLightbox/components/BlockLightboxText/index.js
@@ -20,10 +20,14 @@ export default class BlockLightboxText extends PureComponent {
     mode: 'resting',
   }
 
-  openModal = () => {
+  openModal = e => {
+    //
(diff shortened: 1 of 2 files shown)


src/v2/components/BlockLightbox/components/BlockLightboxText/index.js
```diff
@@ -20,7 +20,15 @@ export default class BlockLightboxText extends PureComponent {
...
-  openModal = () => this.setState({ mode: 'editing' })
+  openModal = () => {
+    if (
+      window.getSelection().toString() ||
+      window.getSelection().toString() !== ''
+    ) {
+      return false
+    }
+    this.setState({ mode: 'editing' })
+  }
```
dzucconi: while ur in here wanna cover link clicks too?

broskoski: sure

PR #2044, opened 2019-05-22
Supports removing owned blocks in other channels

dzucconi:


src/v2/components/ConnectableContextMenu/index.tsx
```diff
@@ -34,7 +41,8 @@ export const ConnectableContextMenu: React.FC<Props> = ({
 
   return (
     <ContextMenu position="absolute" top={8} right={8} zIndex={1} {...rest}>
-      {channel.can.remove_connections && (
+      {(channel.can.remove_connections ||
+        (connectable.__typename !== 'Channel' && connectable.can.remove)) && (
```
broskoski: One should be able to remove a channel they own that is connected to another channel as well

dzucconi: Yeah let's talk about this one...

PR #2059, opened 2019-05-28
Tweaks embed layout

dzucconi:


broskoski:
Should we just go ahead and pull 9 if we can? Or center that 
'+ n more blocks'?

broskoski:
Also, looks like you aren't doing reverse position?

dzucconi:
So: you can adjust per with a query param. It's 8 by default to favor 2-row layouts. Our super narrow blog layout is the exception not the rule.

dzucconi:
That true.

PR #2065, opened 2019-05-29
Limit connect CTA

dzucconi:


src/v2/components/ConnectionSelection/components/ConnectCTA/queries/canConnect.ts
```diff
@@ -0,0 +1,33 @@
...
+      }
+      ... on Block {
+        can {
+          connect
+        }
+      }
+      ... on Channel {
+        can {
+          connect
+        }
```
dzucconi: These will just return `can.connect` => `true` for now until we deploy the ability changes. So this is safe for `master`.

PR #2068, opened 2019-06-03
Add total blocks meter to billing page

broskoski:
Screen Shot 2019-06-03 at 6 39 53 PM

### DIFF
--- a/src/v2/components/TotalBlocksMeter/index.js
+++ b/src/v2/components/TotalBlocksMeter/index.js
@@ -1,0 +1,39 @@
+import React, { Component } from 'react'
+import { propType } from 'graphql-anywhere'
+
+import totalBlocksMeterFragment from 'v2/components/TotalBlocksMeter/fragments/totalBlocksMeter'
+
+import Meter from 'v2/components/UI/Meter'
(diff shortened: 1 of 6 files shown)


src/v2/components/TotalBlocksMeter/index.js
```diff
@@ -0,0 +1,39 @@
...
+import { propType } from 'graphql-anywhere'
+
+import totalBlocksMeterFragment from 'v2/components/TotalBlocksMeter/fragments/totalBlocksMeter'
+
+import Meter from 'v2/components/UI/Meter'
+import Box from 'v2/components/UI/Box'
+
+export default class TotalBlocksMeter extends Component {
+  static propTypes = {
+    me: propType(totalBlocksMeterFragment).isRequired,
```
damassi: There's TypeScript support now, why still JS and using PropTypes?

broskoski: Oops sorry, was moving fast here. This is copy-pasta from another component. Will promise to make all future new components TypeScript.

damassi: 👍

PR #2074, opened 2019-06-05
Move pusher channel connection outside channel contents

broskoski:


broskoski:
I need the value from connecting the channel. Gonna deploy this now, can optimize later.

PR #2087, opened 2019-06-14
Port extension code to bookmarklet

broskoski:
This PR:

1. Moves generic extension code out of extension and into components
2. Extracts extension libs to generic classes, meant to be extended
3. Refactors the way the bookmarklet loader is compiled
4. Adds yarn commands `yarn bookmarklet:build:[dev, prod]` for building `loader.js` file (file that gets added to page when bookmarklet is triggered)


src/apps/marklet/webpack.config.js
```diff
@@ -0,0 +1,65 @@
...
+    }),
+  ],
+  resolve: {
+    extensions: ['.js', '.jsx', '.ts', '.tsx', '.css'],
+    symlinks: false,
+    modules: ['node_modules'],
+  },
+}
+
+module.exports = bookmarkletConfig
```
broskoski: cc @damassi this is what I ended up doing :)

src/apps/marklet/index.js
```diff
@@ -0,0 +1,34 @@
...
+
+  return res.sendFile(asset, options, err => {
+    if (err) {
+      console.log('Error sending bookmarklet:', err)
+      res.status(err.status).end()
+    }
+  })
+}
+
+app.get('/save/:content', (req, res) => res.render('index'))
```
dzucconi: Why is there jade here at all?

src/apps/marklet/templates/index.jade
```diff
@@ -1,11 +1,17 @@
...
+      font-size: 16px !important;
+    }
 
+block container
+  #apolloMount
+    //- rendered on the client
+ 
 block scripts
-  script( src= asset('/assets/marklet.js') )
+  script( src= asset('/assets/bookmarklet.js') )
```
dzucconi: Yeah this template should be unnecessary. You can use the BlankLayout?

src/v2/components/Bookmarklet/components/Login/index.jsx
```diff
@@ -0,0 +1,132 @@
...
+              }
+            </Button>
+          </LoginForm>
+        </CenterStretchBox>
+      </Layout>
+    )
+  }
+}
+
+export default Login
```
dzucconi: So this is just a copy of the existing login form component? What changes?

broskoski: The existing login component is expecting to be rendered server-side, which (server-side rendering) would make this job slightly more complicated to refactor, for no real reason.

src/apps/marklet/index.js
```diff
@@ -24,7 +24,18 @@ const sendAsset = asset => (req, res) => {
...
+    .render(BlankLayout, null, { mode: 'page' })
+    .then(apolloRes => {
+      pageResolver({
+        bundleName: 'bookmarklet',
+        apolloRes,
+        res,
+      })
+    })
+    .catch(next)
+})
```
dzucconi: Bless'd

package.json
```diff
@@ -6,12 +6,14 @@
     "assets": "sh scripts/assets.sh",
     "build": "yarn webpack --config ./webpack/webpack.config.js",
     "build:server": "BUILD_SERVER=true NODE_ENV=production yarn build",
+    "build:bookmarklet:dev": "NODE_ENV=development webpack --config src/apps/marklet/webpack.config.js --watch",
```
damassi: Oh yeah this seems solid 👍

PR #2129, opened 2019-08-07
User settings page 

broskoski:
Converting this page to new stack for secret reasons.


broskoski:
user-settings

broskoski:
Alright this is mergeable. Next steps would be to convert the avatar upload to react, convert the notifications settings, and then convert the whole settings page to a react page.

src/v2/components/UserSettings/index.tsx
```diff
@@ -0,0 +1,382 @@
...
+    me.home_path !== '/explore' &&
+    me.home_path !== `/${me.slug}`
+  const homePath = isCustom ? 'custom' : me.home_path
+
+  const handleSubmit = values => {
+    const variables = {
+      ...values,
+      home_path: values.custom_home_path || values.home_path,
+      custom_home_path: undefined,
+      show_nsfw: values.show_nsfw === 'true',
```
broskoski: This is kind of annoying

src/v2/components/UserSettings/index.tsx
```diff
@@ -0,0 +1,382 @@
+import React from 'react'
+import styled from 'styled-components'
+import { Query, Mutation, MutationFn } from 'react-apollo'
+import { Form, Field } from 'react-final-form'
```
broskoski: Using final-form which is honestly pretty awesome.

src/v2/components/CustomBadgeUploader/index.js
```diff
@@ -4,20 +4,18 @@ import { Query } from 'react-apollo'
...
       <Query query={customBadgeQuery} ssr={false}>
         {({ loading, error, data, startPolling, stopPolling }) => {
           if (loading) {
-            return <LoadingIndicator p={6} />
+            return <span />
           }
 
           if (error) {
-            return <ErrorAlert m={6}>{error.message}</ErrorAlert>
+            return <span />
```
dzucconi: ?

src/v2/components/UserSettings/index.tsx
dzucconi: Well you have the control here; don't have the forms output these values?

src/v2/components/UserSettings/index.tsx
```diff
@@ -0,0 +1,382 @@
...
+  const handleSubmit = values => {
+    const variables = {
+      ...values,
+      home_path: values.custom_home_path || values.home_path,
+      custom_home_path: undefined,
+      show_nsfw: values.show_nsfw === 'true',
+      exclude_from_indexes: values.exclude_from_indexes
+        ? values.exclude_from_indexes === 'true'
+        : undefined,
+    }
```
dzucconi: Yeah I don't follow the why

src/v2/components/UserSettings/index.tsx
dzucconi: What's it give you? I generally don't get why these form libs exist tbh

broskoski: If you’ll notice, it handles all the complex state management. Not just the values but the errors, whether the form has been touched, etc. I don’t enjoy writing that stuff repeatedly when a lib can easily handle the standard patterns. And I don’t reckon I could write it better than someone else who is thinking specifically about forms. 

I originally added it because I wanted a nicer way to handle the custom home path which it handles easily.

src/v2/components/UserSettings/index.tsx
broskoski: These values are coming from the pulldown which I suppose could handle the parsing of strings to bools but it also seems a little weird.

src/v2/components/CustomBadgeUploader/index.js
broskoski: In reality, showing an error here doesn’t really do anything for the user, but we also don’t want to render the component.

src/v2/components/UserSettings/index.tsx
dzucconi: I just think dependencies for simple things like this add more complexity in the long run. Whatever makes you happy, ultimately.

src/v2/components/CustomBadgeUploader/index.js
dzucconi: return null?

src/v2/components/UserSettings/index.tsx
dzucconi: The field should expose correct data; there's nothing to "parse"?

dzucconi: This thing probably shouldn't be a pulldown either IMO? (I know it is one currently)

dzucconi: Like I'd just write a wrapper for the pulldown that just does generic type coercion or something onChange.

broskoski: Ok will do. What would you suggest if not a pulldown? Radio?

dzucconi: Maybe! I think this is just stuck in my head from this morning: https://learnui.design/blog/4-rules-intuitive-ux.html

dzucconi: All the 2-choice options could just be horizontally segmented and I could see a vertically segmented button list with the custom home path split into button + input

broskoski: Point taken. Will revisit.

dzucconi: But you could of course just ship this for now :)

broskoski: No I think you’re right. It doesn’t seem like all that much of a push to switch. At the very least the two button options. I suspect the home path one might need more designing but I’ll give it a think too

broskoski: No I think you’re right. It doesn’t seem like all that much of a push to switch. At the very least the two button options. I suspect the home path one might need more designing but I’ll give it a think too

broskoski:
Screen Shot 2019-08-20 at 2 02 51 PM

dzucconi:
oooo much better!

broskoski:
Gonna merge!

dzucconi:
👍

PR #2147, opened 2019-09-04
Remove cell hover events for mobile devices

broskoski:
Removes weird click behavior for mobile. I welcome alternative approaches.

### DIFF
--- a/src/v2/util/isMobile.ts
+++ b/src/v2/util/isMobile.ts
@@ -1,0 +1,6 @@
+export function isMobile() {
+  return (
+    typeof window.orientation !== 'undefined' ||
+    navigator.userAgent.indexOf('IEMobile') !== -1
+  )
+}
--- a/src/v2/util/isTouchDevice.ts
+++ b/src/v2/util/isTouchDevice.ts
@@ -1,0 +1,4 @@
+// Extracted from https://stackoverflow.com/posts/4819886/revisions
+export function isTouchDevice() {
+  return 'ontouchstart' in window || navigator.maxTouchPoints
+}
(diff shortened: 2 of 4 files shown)


src/v2/util/isMobile.ts
```diff
@@ -0,0 +1,6 @@
+export function isMobile() {
+  return (
+    typeof window.orientation !== 'undefined' ||
+    navigator.userAgent.indexOf('IEMobile') !== -1
```
dzucconi: Shouldn't you be detecting touch?

broskoski: Apparently the presence of  `window.orientation` gets the most coverage for mobile devices. the `IEMobile` is just to catch that Windows mobile devices.

dzucconi: I don't get it though — this is explicitly because of touch devices. It has nothing to do with mobility.

dzucconi: What about touch desktops?

broskoski: Ok, good point

src/v2/util/isTouchDevice.ts
```diff
@@ -0,0 +1,4 @@
+// Extracted from https://stackoverflow.com/posts/4819886/revisions
+export function isTouchDevice() {
+  return 'ontouchstart' in window || navigator.maxTouchPoints
+}
```
dzucconi: https://github.com/aredotna/ervell/blob/master/src/v2/util/is.js#L13

broskoski: Thanks, will remove.

PR #2165, opened 2019-09-07
Use beta styled components

broskoski:
I can only find two styling differences with the new styled components:

Screenshot 2019-09-07 09 59 17
Screenshot 2019-09-07 09 53 48

This appears to be due to how styled components handles `all: initial;` – previously it just set the property, now it appears to actually set all the properties to initial in alphabetical order:

Screenshot 2019-09-07 09 54 19

Therefore, properties like padding get overridden in certain browsers (Chrome).

In any case, removing `all: initial` doesn't seem to cause any noticeable difference (and yes, reset the local server to double check). 

Also added a outlineless mixin for the search bar and the new channel form. Everything else can have outlines.

### DIFF
--- a/package.json
+++ b/package.json
@@ -175,7 +175,7 @@
     "spider-detector": "^1.0.19",
     "sqreen": "^1.30.3",
     "striptags": "^2.2.1",
-    "styled-components": "^4.1.3",
+    "styled-components": "^5.0.0-beta.8",
     "styled-system": "^2.3.2",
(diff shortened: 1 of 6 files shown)


dzucconi:
The use of `all: initial` was to combat the existing unscoped CSS. So I guess that's gone now right?

broskoski:
@dzucconi affirmative.

PR #2181, opened 2019-10-01
Upload via clipboard

broskoski:


src/v2/components/AddBlock/components/AddBlockPasteUploader/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+  channelId,
+}) => {
+  const [mode, setMode] = useState('resting')
+  const [fileUrl, setFileUrl] = useState(null)
+
+  usePasteListener({
+    onPaste: url => {
+      setMode('active')
+      setFileUrl(url)
+    },
```
dzucconi: I'd just avoid inlining callbacks in general and make sure you use useCallback

src/v2/components/AddBlock/components/AddBlockPasteUploader/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+  })
+
+  const finishUpload = () => {
+    setFileUrl(null)
```
dzucconi: ditto

broskoski:
clipboard-paster

src/v2/components/AddBlock/components/AddBlockPasteUploader/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  const onUpload = useCallback(
+    ({ url: value }) => {
+      setFileUrl(null)
+      setMode('resting')
+      return createBlock({ variables: { channel_id: channelId, value } })
+        .then(({ data: { create_block: { block } } }) => onAddBlock(block))
+        .catch(err => {
+          console.error(err)
```
dzucconi: Can we bubble up any kind of user feedback? It's unfortunate we don't have a toast-like error UI we could use.

src/v2/hooks/usePasteListener/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+      transformImages(e.clipboardData, acceptedFiles, (url: string) => {
+        setFileUrl(url)
+        onPaste(url)
+      })
+    }
+
+    el.addEventListener('paste', pasteHandler)
+
+    return () => {
+      el.removeEventListener('paste', pasteHandler)
```
dzucconi: ✔️

src/v2/components/AddBlock/components/AddBlockPasteUploader/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+  right: 0;
+  bottom: 0;
+  left: 0;
+  z-index: 3;
+`
+
+interface AddBlockPasteUploader {
+  onAddBlock: (props: any) => any
+  channelId: string | number
+  createBlock: (props: any) => Promise<any>
```
dzucconi: Would avoid all these anys. I know there's a lot of them but just get rid of them. Hover to find out return types at the call sites and just paste em around.

src/v2/components/AddBlock/components/AddBlockPasteUploader/index.tsx
broskoski: Yeah, true. Let me see what you do in the drop zone uploader.

src/v2/components/AddBlock/components/AddBlockPasteUploader/index.tsx
broskoski: Now I remember why I did that:
Screen Shot 2019-10-07 at 1 34 40 PM

Will fix.

dzucconi: Yeah, I suspect most of my TypeScript from this era isn't exactly exemplar.

PR #2202, opened 2019-11-26
Fix cursor on Delete Channel button hover

damassi:
Since the `a` tag doesn't have an `href` it needed extra styling.


broskoski:
@damassi you honor us with your code

PR #2205, opened 2019-12-05
Convert settings app to full react

broskoski:
settings-page


src/v2/components/UI/Page/index.js
```diff
@@ -154,6 +157,7 @@ export default class Page extends PureComponent {
             type="module"
             integrity="sha384-2xV8M5griQmzyiY3CDqh1dn4z3llDVqZDqzjzcY+jCBCk/a5fXJmuZ/40JJAPeoU"
           />
+          {includeStripe && <script src="https://js.stripe.com/v3/" />}
```
dzucconi: You should add this using Helmet

broskoski: Ah yes, good call

PR #2222, opened 2020-01-08
Upgrade apollo to 3.1

broskoski:


src/extension/src/apollo/index.tsx
```diff
@@ -50,14 +38,23 @@ export const initApolloClient = ({
...
   })
 
+  const data = {
+    loginStatus: {
+      __typename: 'LoginStatus',
+      isLoggedIn,
+    },
+  }
+
+  cache.writeData({ data })
```
damassi: This is nicer

src/v2/components/AddBlock/components/AddBlockPasteUploader/index.tsx
```diff
@@ -1,6 +1,6 @@
 import React, { useState, useCallback } from 'react'
 import styled from 'styled-components'
-import { MutationFn } from 'react-apollo'
+import { MutationFunction as MutationFn } from 'react-apollo'
```
damassi: Probably already aware but heads up about the new `@apollo/react-hooks` package: 
```js
import { useQuery, useLazyQuery } from "@apollo/react-hooks"
```

broskoski: Ya, that's mostly why I'm upgrading :)

PR #2223, opened 2020-01-08
Upgrade typescript@3.7.3, upgrade Babel and add optional chaining plugin

damassi:
This bumps typescript to the latest so that two recently-advanced ECMA features could be brought in via babel: optional chaining (`foo?.bar?.baz`) and nullish coalescing (`foo ?? bar`). See the [latest TypeScript release notes](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining) for more info. 

Also bumps `@babel` packages to latest.


broskoski:
Thank you!!!

PR #2235, opened 2020-01-09
Refactor commenting using react-mentions

broskoski:
So APPARENTLY! doing triggers on `keyCode` is unreliable, as different keyboards map the @ symbol to different keyCodes. 

This PR refactors mentions to use the `react-mention` library (which is much better anyway) that uses the actual character rather than the code.


broskoski:
Screen Shot 2020-01-09 at 2 55 54 PM

src/v2/components/UI/MentionTextarea/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+  onChange,
+}) => {
+  const [getSuggestions, { refetch }] = useLazyQuery<
+    MentionTextareaUserSuggestions
+  >(USER_SUGGESTION_QUERY)
+
+  const fetchSuggestions = (query: string, callback: () => void) => {
+    return refetch({ query })
+      .then(results => {
+        if (!results?.data?.suggestions) return []
```
damassi: yayy

PR #2238, opened 2020-01-10
Add cursor pointer on arrow hover

damassi:
Another silly small one -- adds a mouse cursor pointer when hovering (only) over the SVG arrow.

Screen Shot 2020-01-09 at 11 51 01 PM


broskoski:
Thank you!

PR #2244, opened 2020-01-16
Add block and channel counts to filter bar

broskoski:
Screen Shot 2020-01-16 at 5 24 50 PM

### DIFF
--- a/src/v2/components/ChannelContentsFilter/index.tsx
+++ b/src/v2/components/ChannelContentsFilter/index.tsx
@@ -2,7 +2,6 @@ import React, { useState } from 'react'
 import { Query } from 'react-apollo'
 import styled from 'styled-components'
 import { unescape } from 'underscore'
-import { toSentence } from 'underscore.string'
 
 import constants from 'v2/styles/constants'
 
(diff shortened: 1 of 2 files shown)


src/v2/components/ChannelContentsFilter/index.tsx
```diff
@@ -39,11 +40,24 @@ export const ChannelContentsFilter = ({ channel }) => {
...
+  const channelLabel =
+    channel.counts.channels === 1
+      ? '1 channel'
+      : `${channel.counts.channels} channels`
+  const blockLabel =
+    channel.counts.blocks === 1 ? '1 block' : `${channel.counts.blocks} blocks`
+
+  const countLabels = []
+  if (channel.counts.blocks > 0) countLabels.push(blockLabel)
+  if (channel.counts.channels > 0) countLabels.push(channelLabel)
```
dzucconi: I'm the kinda person that would find something like

```javascript
const { counts } = channel;

const counts = toSentence([
  ...(counts.channels > 0 ? [`${counts.channels} channel${counts.channels === 1 ? '' : 's'}`] : []),
  ...(counts.blocks > 0 ? [`${counts.blocks} block${counts.blocks === 1 ? '' : 's'}`] : []),
]);
```

to be more legible.

Tho I'd probably just pull out some different helpers here and try to get rid of underscore.string

So this is fine.

broskoski: Yeah, I knew you'd have opinions on this.

I think your version is nicer and yeah I would like to avoid `toSentence` so let me try to rework.

broskoski:
There we go...
Screen Shot 2020-01-16 at 5 43 02 PM
Screen Shot 2020-01-16 at 5 43 14 PM

dzucconi:
🍏

PR #2253, opened 2020-01-24
[Feature] Add draggable SplitPane support to BlockLightbox

damassi:
This PR adds draggable split-pane support to the `BlockLightbox` component, via [`react-split-pane`](https://github.com/tomkp/react-split-pane). The user can now resize the description / main content area as they see fit. The divider only appears when the user hovers in that center area, and since it's near the prev / next arrows I'm assuming it will be discoverable. (It's also a common idiom to resize panes like this, so perhaps users will just know its there.)

The idea for this came from me often having multiple browser windows open side by side on my laptop, and at smaller dimensions it became difficult to read the details pane. I wanted the ability to drag it out. Also, my thinking that this description panel could grow into something more full-featured. 

There are a few hairy spots in the code, commented inline. 

resize


src/v2/components/BlockLightbox/components/BlockLightboxContentPane/index.tsx
```diff
@@ -1,9 +1,7 @@
 import React, { PureComponent } from 'react'
 import styled from 'styled-components'
-import PropTypes from 'prop-types'
-import { propType } from 'graphql-anywhere'
 
-import blockLightboxContentPaneFragment from 'v2/components/BlockLightbox/components/BlockLightboxContentPane/fragments/blockLightboxContentPane'
+import { BlockLightbox as Block } from '__generated__/BlockLightbox'
```
damassi: The few .js files I touched I converted to TypeScript

src/v2/components/BlockLightbox/components/BlockLightboxMetadataPane/index.tsx
```diff
@@ -45,6 +48,14 @@ export default class BlockLightboxMetadataPane extends PureComponent {
...
   }
 
+  get canManage() {
+    if (this.props.block.__typename !== 'Channel') {
+      return this.props.block.can.manage
+    } else {
+      return false
+    }
+  }
+
```
damassi: When converting to .ts, noticed this type-error -- `can.manage` doesn't exist on `Channel`, so this makes things a bit safer.

src/v2/components/ModalBlockLightbox/index.js
```diff
@@ -21,6 +21,7 @@ import ModalBlockLightboxNavigation from 'v2/components/ModalBlockLightbox/compo
 const Fullscreen = styled(Link).attrs({
   border: '1px solid',
 })`
+  cursor: pointer;
```
damassi: Fix missing cursor on full-screen button

src/v2/components/UI/SplitPane/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+      flex-direction: column !important;
+      overflow: scroll !important;
+
+      .Pane1,
+      .Pane2 {
+        width: 100% !important;
+      }
+      .Resizer {
+        display: none !important;
+      }
```
damassi: Sorry for these `!importants`. I couldn't figure out how to "disable" `react-split-pane` any other way. When first rendering we can lean on the `isMobile` function helper, but if the user resizes the browser or rotates or whatever then we have to forcefully use CSS, via a media query, to return the display to its mobile form. Since `react-split-pane` uses inline styles to apply behavior, `!important` was the only way to override. 

A PR upstreram with a `disable` prop would fix this. Then, if true, it returns a pass-through component rather than initializing functionality. 

Happy for suggestions on how to fix.

src/v2/components/UI/SplitPane/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    &:after {
+      border-left: 1px solid ${props => props.theme.colors.gray.light};
+      content: '';
+      display: flex;
+      position: relative;
+      width: 1px;
+      height: 100%;
+    }
```
damassi: Center the vertical rule within the hit area

src/v2/components/UI/SplitPane/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+    >
+      <ReactSplitPane
+        split="vertical"
+        primary={'second'}
+        allowResize={!isFullScreen}
+        ref={splitPane}
+        defaultSize={panelSize || minSize}
+        minSize={minSize}
+        size={isFullScreen ? '0%' : sizeAfterUpdate}
+        step={5}
```
damassi: Slight perf boost by adjusting step

src/v2/hocs/WithIsSpiderRequesting/queries/isSpiderRequesting.ts
```diff
@@ -3,6 +3,7 @@ import gql from 'graphql-tag'
 export default gql`
   query IsSpiderRequestingQuery {
     sharify @client {
+      # FIXME: Where is this constant coming from?
```
damassi: Noticing this Apollo validation error and wondering if this is a bug?

src/v2/util/isMobile.ts
```diff
@@ -1,4 +1,8 @@
 export function isMobile() {
+  if (typeof window === 'undefined') {
+    return false
+  }
```
damassi: Guard during SSR

src/v2/components/BlockLightbox/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+import constants from 'v2/styles/constants'
+
+import { BlockLightbox as Block } from '__generated__/BlockLightbox'
+
+import Box from 'v2/components/UI/Box'
+import Title from 'v2/components/UI/Head/components/Title'
+import BlockLightboxContentPane from 'v2/components/BlockLightbox/components/BlockLightboxContentPane'
+import BlockLightboxMetadataPane from 'v2/components/BlockLightbox/components/BlockLightboxMetadataPane'
+import { SplitPane } from 'v2/components/UI/SplitPane'
```
damassi: Just a note that avoiding default exports helps contain drift. With a default export, the person importing can name the variable whatever, where a named export ties things 1:1 to the object being exported, lending itself to static analysis, find and replace, maintenance over time, etc.

src/v2/components/BlockLightbox/components/BlockLightboxMetadataPane/index.tsx
dzucconi: Hm, this _should_ definitely be there on Channel always

src/v2/components/BlockLightbox/index.tsx
dzucconi: Yeah, I've completely moved over personally as well.

src/v2/components/UI/SplitPane/index.tsx
```diff
@@ -0,0 +1,173 @@
...
+
+  .Pane2 {
+    min-width: ${props => (props.isFullScreen ? 0 : MIN_PANEL_WIDTH.px)};
+  }
+
+  .Resizer {
+    cursor: ${props => (props.isFullScreen ? 'inherit' : 'col-resize')};
+    position: ${props => (props.isFullScreen ? 'absolute' : 'relative')};
+    padding-left: 10px;
+    padding-right: 10px;
```
dzucconi: I think we have a space scale in the theme

src/v2/hocs/WithIsSpiderRequesting/queries/isSpiderRequesting.ts
dzucconi: ?

dzucconi:
Does this set a local preference for the width?

dzucconi:
I've never had the need to do this, but I'm not opposed to it

src/v2/components/BlockLightbox/components/BlockLightboxMetadataPane/index.tsx
broskoski: Yeah, it definitely should be there. I'll take a look..

src/v2/hocs/WithIsSpiderRequesting/queries/isSpiderRequesting.ts
broskoski: It's coming from here: https://github.com/aredotna/ervell/blob/master/src/lib/middleware/isSpider.js

src/v2/components/BlockLightbox/index.tsx
broskoski: Will do, thanks for that info. I thought it was just a style thing.

broskoski:
Tested locally, looks good on desktop and feels good to adjust. Very nice 👍 

On mobile though, I'm not seeing any content other than the sidebar:
Screen Shot 2020-01-24 at 12 38 14 PM

src/v2/components/BlockLightbox/components/BlockLightboxMetadataPane/index.tsx
damassi: Cool, i'll revert this change then 👍

damassi:
> Does this set a local preference for the width?

@dzucconi - No it does not. It seemed like a bit of overkill for now but I can certainly add it if you think it's important.

damassi:
@broskoski - what browser / phone? Everything looks good over here on iOS. I can check it out on browserstack though. 

Screen Shot 2020-01-24 at 10 38 04 AM

broskoski:
Ah, interesting. I was testing on Safari in responsive mode.

damassi:
Ah, I didn't check Safari's responsivity -- seems like that's the bug. Fixin...

src/v2/hocs/WithIsSpiderRequesting/queries/isSpiderRequesting.ts
damassi: Fixed by adding a type to [here](https://github.com/aredotna/ervell/blob/master/src/v2/apollo/localState/clientSchema.graphql#L41-L43)

src/v2/components/BlockLightbox/components/BlockLightboxMetadataPane/index.tsx
damassi: https://github.com/aredotna/ervell/pull/2253/commits/ddbbac7c55e97418dcbbb1bbc05e469f48b9a1f7

damassi:
This is ready for another look

broskoski:
Looks good to me, thank you @damassi ❤️

PR #2255, opened 2020-01-25
[Feature] Add DarkMode toggle to User Settings

damassi:
Adds the ability to explicitly set dark mode via the user's settings page making things a bit more discoverable. 

Not sure how to [update the docs](https://support.are.na/help/is-there-a-dark-mode), though, which only mention the key combo. 

darkmode

### DIFF
--- a/src/v2/components/UserSettings/index.tsx
+++ b/src/v2/components/UserSettings/index.tsx
@@ -21,6 +21,7 @@ import Text from 'v2/components/UI/Text'
 import RadioOptions from 'v2/components/UI/RadioOptions'
 import Pulldown from 'v2/components/UI/Pulldown'
 import HorizontalRule from 'v2/components/UI/HorizontalRule'
+import { toggle as toggleOrSetTheme } from 'v2/components/UI/Layouts/BlankLayout/components/LegacyDarkTheme'
 
 import mapErrors from 'v2/util/mapErrors'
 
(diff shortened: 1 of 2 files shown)


damassi:
One q I have is would it make sense to also place a toggle in the main shell or directly in the nav? When browsing through content its nice to be able to quickly toggle between things for a/b purposes because each mode makes different visual qualities pop out. That said being under settings does match the mobile app.

broskoski:
I think it's fine to leave here now. I can imagine putting it in the main menu, but this is so straightforward and like you said, mirrors the mobile app.

src/v2/components/UserSettings/index.tsx
```diff
@@ -480,6 +481,47 @@ const UserSettings: React.FC<UserSettingsProps> = ({ me, updateAccount }) => {
...
+
+              <Field
+                name="darkmode"
+                initialValue={document.documentElement.dataset.theme === 'dark'}
+              >
+                {props => {
+                  return (
+                    <>
+                      <InputContainer mt={6} mb={5}>
+                        <Label>Darkmode</Label>
```
broskoski: So minor: this should be "Dark mode"

src/v2/components/UserSettings/index.tsx
```diff
@@ -480,6 +481,47 @@ const UserSettings: React.FC<UserSettingsProps> = ({ me, updateAccount }) => {
...
+                          value={props.input.value}
+                          onSelect={value => {
+                            const theme = value === true ? 'dark' : 'default'
+                            toggleOrSetTheme(theme)
+                          }}
+                        >
+                          <RadioOptions.Option value={true}>
+                            {({ selected }) => (
+                              <Text f={4} mb={3} selected={selected}>
+                                Enable Darkmode
```
broskoski: Also minor: change to just "Enable / Disable"

damassi: I set it to be more verbose here based on the other toggles, which explicitly spell out each option. Assumed it was related to accessibility. Will fix 👍

PR #2257, opened 2020-02-06
Safari extension

broskoski:
So there are still some things to fix here, but I'm curious if this approach is acceptable. It's basically the same issue as the other (Chrome, Firefox) browser extensions (we want the compiled component code to be inside the packaged extension) but more gross because there would now be Swift code inside of Ervell as well.

Happy to take suggestions here. One option is to make it a git submodule...


package.json
```diff
@@ -8,6 +8,8 @@
     "build:server": "BUILD_SERVER=true NODE_ENV=production yarn build",
     "build:bookmarklet:dev": "NODE_ENV=development webpack --config src/apps/marklet/webpack.config.js --watch",
     "build:bookmarklet:prod": "NODE_ENV=production webpack --config src/apps/marklet/webpack.config.js",
+    "build:safari-ext:dev": "NODE_ENV=production webpack --config src/extension/safari-extension/webpack.config.js --watch",
```
damassi: Should `NODE_ENV=production` be set for the `safari-ext:dev` script?

broskoski: No, good catch

src/extension/safari-extension/Are.na for Safari/Are.na for Safari Extension/SafariExtensionViewController.swift
```diff
@@ -0,0 +1,342 @@
...
+        let json = (jsonSerialize(obj: message) ?? "null")
+        webView.evaluateJavaScript("window.arenaSafariAppMessageReceiver(\(json));", completionHandler: nil)
+    }
+    
+    func sendCurrentPage() {
+        NSLog("Getting current page")
+        let options = TabQueryOptions()
+        options.active = true
+        SFSafariApplication.getActiveWindow { win in
+            if ((win) != nil) {
```
damassi: #minor `(win)`

src/extension/safari-extension/background.ts
```diff
@@ -0,0 +1,138 @@
+export class SafariApp {
```
damassi: Can short circuit the window casting via a 
```ts
declare const window: any
```
At the top of the file

damassi:
All I gotta say @broskoski is 😅😅😅😅(Also props on figuring this out)

Re keeping Swift code in Ervell repo: it's common in large apps to have different languages mixed together. Unless there are huge binaries checked in I'm 👍Submodules are a pain.

package.json
broskoski: Actually, easiest to leave this on `production` (i.e. it's going to use the graphql production api endpoint)

broskoski:
Merging, thank you!

PR #2260, opened 2020-02-10
Silence type generation errors

broskoski:
@damassi fixed most of the errors instead of the main one, which is:

```
Error: Cannot use GraphQLScalarType "String" from another module or realm.
Ensure that there is only one instance of "graphql" in the node_modules
directory. If different versions of "graphql" are the dependencies of other
relied on modules, use "resolutions" to ensure only one version is installed.
https://yarnpkg.com/en/docs/selective-version-resolutions
Duplicate "graphql" modules cannot be used at the same time since different
versions may have different capabilities and behavior. The data from one
version used in the function from another could produce confusing and
spurious results.
```


damassi:
👍 Let me see if i can fix the main issue in a separate PR and can rebase into this one.

damassi:
Try this: https://github.com/aredotna/ervell/pull/2261

PR #2261, opened 2020-02-10
Bump graphql@14.5.8

damassi:
This bumps and locks things, but i think what it came down to was 
```
rm -rf node_modules && yarn --check-files
```
/shrug


broskoski:
Ah it was probably the `@types`

broskoski:
Yep, it works!

PR #2262, opened 2020-02-10
Save selected channels in extension state

broskoski:
This is a very subtle bug but required a bit of refactoring. We basically want a way to pass through selected channels through the connect component so that we can retain state on the default screen.

The bug report was:

```

1. Select Image
2. Right Click > Add to are.na
3. Type in channel name
4. Click Channel name [button on below changes to save to 1 channel]
5. Edit image name / description
6. Click Save
7. Search the channel again(because it disappeared), now you can click on the channel name again and below it says [save to 2 channels] although that channel was already selected.
8. If you keep clicking on the same channel name, you can increase the number below where it says [save to xx channels] 

```

**Currently in production:**

safari-extension-post

**After this PR:**

safari-extension-pre

Alongside this, I've refactored all the components I touched into TypeScript and needed to clear up some warnings so there are some new generated files in here as well.


src/v2/components/ConnectionSelectionList/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+export type onConnectionSelectionType = (
+  isSelected: boolean,
+  channel: Channel
+) => void
+
+export interface ConnectionSelectionListProps {
+  isOutlined?: boolean
+  onConnectionSelection?: onConnectionSelectionType
+  selectedChannels?: Channel[]
```
broskoski: First, allow `selectedChannels` to be passed down

src/v2/components/ConnectionSelectionList/components/RecentChannels/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+  const mergedChannels = [...selectedChannels, ...data.me.recent_channels]
+
+  const channels = mergedChannels.reduce(
+    (channels, channel) =>
+      channels.find(x => x.id === channel.id)
+        ? [...channels]
+        : [...channels, channel],
+    []
+  )
```
broskoski: Merge recent channels with selected channels and filter out the duplicates

src/v2/components/ConnectionSelectionList/components/RecentChannels/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+  return (
+    <>
+      <ChannelsList
+        channels={channels}
+        selectedChannels={selectedChannels}
```
broskoski: Also pass the original selected channels list

src/v2/components/ConnectionSelectionList/components/ChannelsList/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+  onConnectionSelection,
+  selectedChannels = [],
+  ...rest
+}) => {
+  return (
+    <Container {...rest}>
+      {channels.length === 0 && <ListButton disabled>Nothing yet.</ListButton>}
+
+      {channels.map(channel => {
+        const isSelected = selectedChannels.find(c => c.id === channel.id)
```
broskoski: Find the current channel in the selected channels list, and mark it

src/v2/components/ConnectionSelectionList/components/SelectableChannel/index.tsx
```diff
@@ -119,76 +119,60 @@ const GroupBadge = styled(Badge)`
...
+}
 
-    this.setState(({ isSelected }) => {
-      onSelection(!isSelected, channel.id)
+export const SelectableChannel: React.FC<SelectableChannelProps> = ({
+  channel,
+  onSelection,
+  isSelected,
+}) => {
+  const [selected, setSelected] = useState(isSelected)
```
broskoski: Use the value of whatever the `isSelected` prop is to set the state.

broskoski:
I think the cool side-effect that this has is that it's now easy to preload `selectedChannels` to the connect component. Combining this with the `your_connections` query will make it so that we can preselect already connected channels into the connect component (this is more for in the actual application than the extension).

src/v2/components/ConnectionSelectionList/components/ChannelsList/index.tsx
```diff
@@ -0,0 +1,42 @@
+import React from 'react'
```
damassi: Thanks for converting touched stuff to TS 👍

src/v2/components/ConnectionSelectionList/components/RecentChannels/index.tsx
damassi: ^ would be a nice comment right there in the actual code

src/v2/components/ConnectionSelectionList/components/RecentChannels/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+  const channels = mergedChannels.reduce(
+    (channels, channel) =>
+      channels.find(x => x.id === channel.id)
+        ? [...channels]
+        : [...channels, channel],
+    []
+  )
+
+  return (
+    <>
```
damassi: Lil nit, no need for fragment wrapper

src/v2/components/ConnectionSelectionList/components/RecentChannels/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+  if (!data) {
+    return null
+  }
+
+  const mergedChannels = [...selectedChannels, ...data.me.recent_channels]
+
+  const channels = mergedChannels.reduce(
+    (channels, channel) =>
+      channels.find(x => x.id === channel.id)
+        ? [...channels]
```
damassi: `? channels`

src/v2/components/ConnectionSelectionList/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+  ${x =>
+    x.isOutlined &&
+    `
+    &:after {
+      ${__outlineBorder__()}
+    }
+  `}
+`
+
+export type onConnectionSelectionType = (
```
damassi: Generally types and interfaces are caps-case: `OnConnectionSelection`

broskoski:
Thanks @damassi !!

PR #2267, opened 2020-02-13
[Feature] Add `/` hotkey to focus global search

damassi:
This adds a new mousetrap `/` hotkey so that it's easy to jump to global search without using a mouse.

### DIFF
--- a/src/v2/components/TopBar/components/PrimarySearch/index.js
+++ b/src/v2/components/TopBar/components/PrimarySearch/index.js
@@ -106,6 +106,7 @@ export default class PrimarySearch extends PureComponent {
         {mode === 'resting' && <HomeLink />}
 
         <SearchInput
+          globallyFocusOnKey="/"
           tabIndex={1}
           flex="1"
           py={6}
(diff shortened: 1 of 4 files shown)


broskoski:
> When inside channel view / will jump to channel filter search, taking precedence. Would be best for behavior to be consistent, or is this a nice side-effect?

Hm, I think I would expect that this should always be consistent but I can also see arguments the other way. One argument for keeping it consistent: if it triggered the filter bar, and you were far down in a channel it would trigger a jump to the filter bar scroll position (right?)

damassi:
I believe so, yup -- runs `focus()` on the element. I def think consistency is better, but also find the global search bar more useful at least in terms of my own use patterns, quickly jumping between channels.

src/v2/components/TopBar/components/PrimarySearch/index.js
```diff
@@ -106,6 +106,7 @@ export default class PrimarySearch extends PureComponent {
         {mode === 'resting' && <HomeLink />}
 
         <SearchInput
+          globallyFocusOnKey="/"
```
damassi: Added this setting, which should allow us to be able to add additional commands later to nested search inputs (like in channels).

damassi:
Aight @broskoski this is ready to go

broskoski:
Thank you @damassi !

PR #2274, opened 2020-02-25
Fix NAN around cover image edit date

damassi:
Noticed this with MP3 blocks: 

Screen Shot 2020-02-24 at 10 03 54 PM


broskoski:
You're an angel

PR #2283, opened 2020-03-17
Edit channel modal on channel index

broskoski:
edit-button


src/v2/components/CompactChannel/index.tsx
```diff
@@ -0,0 +1,173 @@
...
+          <Button color={`channel.${channel.visibility}`} onClick={onEditClick}>
+            Edit
+          </Button>
+        </Label>
+      )}
+    </Container>
+  )
+}
+
+export const MemoizedCompactChannel = React.memo(CompactChannel)
```
dzucconi: Why not just wrap compact channel directly, why the namespacing?

broskoski: mm sure, can do

src/v2/components/ManageChannel/index.tsx
```diff
@@ -0,0 +1,349 @@
...
+      }}
+    />
+  )
+}
+
+interface ManageChannelContainerProps {
+  id: string | number
+  refetchQueries?: PureQueryOptions[]
+  onUpdate?: () => any
+  onDelete?: () => any
```
dzucconi: generally you want to return void if you're ignoring the callback's return value

PR #2288, opened 2020-03-25
Re-invert iframe in dark mode

broskoski:
Before
Screen Shot 2020-03-25 at 4 58 21 PM

After
Screen Shot 2020-03-25 at 4 55 58 PM


src/v2/components/UI/Layouts/BlankLayout/components/LegacyDarkTheme/index.js
```diff
@@ -18,6 +18,10 @@ export default createGlobalStyle`
     canvas,
     [style*='background-image'] {
       filter: invert(100%) hue-rotate(180deg);
+
+      iframe {
+        filter: invert(100%) hue-rotate(180deg);
+      }
     }
```
dzucconi: That'll do it!

PR #2294, opened 2020-04-04
SPA

dzucconi:
So not super together. The block modal is one issue that needs to be dealt with. We can hack around it by just making CompactChannel normal a#hrefs in that context or something. Or just fix it. Gotta pull over explore too, minor thing. WIll pick this back up in a bit; but checkout the branch and take a look.


.yarnrc.yml
```diff
@@ -0,0 +1 @@
+yarnPath: .yarn/releases/yarn-1.22.4.js
```
dzucconi, line 1: I run Yarn 2 so this just checks in 1.x. That's what most of this diff is in line count.

dzucconi:
But basically; just to be clear; all we have to do is: pull over the routes into a single switch and update the links to be react-router links. So really not much.

src/apps/app/App.ts
```diff
@@ -0,0 +1,23 @@
...
+    req.apollo
+      .render(withStaticRouter(Routes), null, { mode: 'page' })
+      .then(apolloRes => {
+        pageResolver({ bundleName: 'app', apolloRes, res })
+      })
+      .catch(next)
+  },
+]
+
+App.get('*', ...resolve)
```
dzucconi, line 23: **

src/apps/app/Routes.tsx
```diff
@@ -0,0 +1,80 @@
...
+        path="/block/:id"
+        render={parseRoute(({ params }) => (
+          <BlockPage id={parseInt(params.id, 10)} />
+        ))}
+      />
+
+      {/* Profile */}
+      <Route
+        exact
+        path="/:id/:view(all|index|blocks|channels|followers|following|groups)?"
```
dzucconi, line 40: This is how we get around the ambiguous routing issue here. It'll only match the profile routes if it's constrained by one of these reserved words, otherwise falls through to the channel.

src/apps/index.coffee
```diff
@@ -4,108 +4,114 @@ express = require 'express'
...
-
-console.time('rss')
-app.use(require('../apps/rss'))
-console.timeEnd('rss')
-
-# Ennvironment specific
-switch NODE_ENV
-  when 'development'
-    app.use require '../apps/statuses'
+# console.time('actions')
```
broskoski, line 7: What happens when these are uncommented?

src/apps/profile/Routes.js
```diff
@@ -1,78 +1,14 @@
...
-import { Query } from 'react-apollo'
 import { Switch, Route } from 'react-router-dom'
-
-import profileUiStateQuery from 'apps/profile/queries/profileUiState'
-
 import parseRoute from 'v2/util/parseRoute'
-
 import ProfilePage from 'v2/pages/profile/ProfilePage'
 
-const VALID_SORTS = ['UPDATED_AT', 'RANDOM']
```
broskoski, line 11: Why delete this?

src/v2/pages/profile/ProfilePage/index.js
```diff
@@ -130,3 +130,67 @@ export default class ProfilePage extends Component {
     )
   }
 }
+
+import profileUiStateQuery from 'apps/profile/queries/profileUiState'
```
broskoski, line 134: Nevermind, I get it

src/apps/index.coffee
dzucconi: They'd match first; we'll just have to remove any of the routes that match here then fall thru to our wildcard.

src/apps/profile/Routes.js
dzucconi: I just moved to the page itself.

src/apps/index.coffee
broskoski: Or position the wildcard route up a bit?

dzucconi: Nah, we'll uncomment the one's that are needed is all. Don't think of this as an actual change here for now.

broskoski:
So apologies for the gigantic PR but this is essentially ready. There are probably some edge-cases I am missing, but for the most part its pretty straightforward.

The biggest thing was getting the filter states to persist, which I know is a stupid feature but it just bugged me. Everything else was mostly just adding in the routes and cleaning up some pages / converting to ts.

dzucconi:
OK Cool, so to make this mergeable we just gotta uncomment all the old apps that are still outside the SPA?

broskoski:
@dzucconi already done. I _think_ it's mergeable, at least to test on staging.

broskoski:
@dzucconi did not do any kind of code splitting or anything fancy like that. This is really just making the client-side routes work as expected.

dzucconi:
huh?

broskoski:
Oi, must have happened when I resolved conflicts, just a sec

PR #2295, opened 2020-04-06
Get lightbox almost working

broskoski:
lightbox


damassi:
@broskoski - curious, what does this PR do exactly? Was thinking that it updated URL bar on box open but prod already does that?

damassi:
Ahh SPA work -- right on 💯

PR #2315, opened 2020-04-28
Add isSpiderRequesting hook and fix for SSR rendered pages

broskoski:


src/v2/components/Cell/components/Konnectable/index.tsx
```diff
@@ -16,6 +16,8 @@ import { KonnectableDisplay } from 'v2/components/Cell/components/Konnectable/co
 import KonnectableMetadata from 'v2/components/Cell/components/Konnectable/components/KonnectableMetadata'
 import KonnectableBlockOverlay from 'v2/components/Cell/components/Konnectable/components/KonnectableBlockOverlay'
 import KonnectableChannelOverlay from 'v2/components/Cell/components/Konnectable/components/KonnectableChannelOverlay'
+import useIsSpiderRequesting from 'v2/hooks/useIsSpiderRequesting'
+import isSpiderRequesting from 'v2/hocs/WithIsSpiderRequesting/queries/isSpiderRequesting'
```
dzucconi, line 20: Wait, this is importing a query? It's hard to tell what's going on cuz the variable is named the same

broskoski: Oh weird, will fix.

PR #2326, opened 2020-05-04
Adds breadcrumb loader to links

broskoski:
spa-load

There are a lot of changes from both types being regenerated and new files being converted to ts.

But the basic idea is, for whatever object we want to link to: 
- fetch the required segments
- pass those segments as state through the router
- loading page picks up the state and displays the breadcrumbs if available.


src/v2/util/getBreadcrumbPath.ts
```diff
@@ -0,0 +1,34 @@
+import { SerializeMeQueryHook_serializedMe } from '__generated__/SerializeMeQueryHook'
+import { LoadingBreadcrumbChannel } from '__generated__/LoadingBreadcrumbChannel'
+import { LoadingBreadcrumbUser } from '__generated__/LoadingBreadcrumbUser'
+import { LoadingBreadcrumbGroup } from '__generated__/LoadingBreadcrumbGroup'
+
+type Crumbable =
```
dzucconi, line 6: lol this name

dzucconi: vaguely disappointed it wasn't named `type Crumby`

broskoski: I can do that for you

dzucconi:
Looks good, pleased about this change which I think will make a big difference in perception.

PR #2336, opened 2020-05-14
Update text block and manage block

broskoski:
text-block
Screen Shot 2020-05-14 at 3 34 49 PM


src/v2/components/UI/SansSerifText/index.tsx
```diff
@@ -0,0 +1,179 @@
...
+  font-size: 13px;
+
+  line-height: 1.45;
+
+  p,
+  li,
+  ol,
+  code,
+  pre {
+    font-size: 0.9rem / em;
```
dzucconi, line 30: I've literally never seen this syntax before; what's this division thing?

PR #2360, opened 2020-06-05
Link reader mode

broskoski:
(Ignore the line count, it's mostly the schema being regenerated)

This launches a _really_ rough version of reader mode for lifetime premium & supporters only. Depends on https://github.com/aredotna/api/pull/1190.

Screen Shot 2020-06-05 at 2 56 19 PM


src/v2/components/BlockLightbox/components/BlockLightboxLink/components/BlockLightboxLinkScreenshot/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+import { BlockLightboxLinkProps } from 'v2/components/BlockLightbox/components/BlockLightboxLink'
+
+const Screenshot = styled(Box)``
+
+const Image = styled(Box)`
+  position: relative;
+  background-size: cover;
+  background-position: top center;
+  background-image: url(${props => props.backgroundImage});
```
dzucconi, line 16: I'd set this with `style` I'm sure we've messed this up in other places but generally consider that styled-components generates classes, so it would generate a new class for every single possible input on a page here. In this case there's only one on screen at a time so... nbd. But *in general* only pass props where you've got a finite set of possibilities.

src/v2/components/BlockLightbox/components/BlockLightboxLink/components/BlockLightboxLinkScreenshot/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+  &:after {
+    content: '';
+    display: block;
+    position: absolute;
+    right: 0;
+    bottom: 0;
+    left: 0;
+    height: 6em;
+    background: linear-gradient(
+      ${props => props.theme.colors.utility.transparent} 0%,
```
dzucconi, line 27: At some point u might want to install the `themeGet` function and just use that which would cut down on the tedium of these things `linear-gradient(${themeGet('colors.utility.transparent')})`

src/v2/components/BlockLightbox/components/BlockLightboxLink/index.tsx
```diff
@@ -0,0 +1,26 @@
...
+  layout: LightboxLayout
+  block: Block
+  linkViewMode: LinkViewMode
+}
+
+export const BlockLightboxLink: React.FC<BlockLightboxLinkProps> = ({
+  linkViewMode,
+  ...rest
+}) => {
+  console.log('BlockLightboxLink', { linkViewMode })
```
dzucconi, line 19: 👋

src/v2/components/BlockLightbox/components/BlockLightboxLink/components/BlockLightboxLinkScreenshot/index.tsx
broskoski: Ah got it. Will fix that. This is mostly just a renaming of this file, to .tsx, but it's a good catch.

PR #2380, opened 2020-08-19
Blog rewrite to React / GraphQL

broskoski:
rewritten-blog

The only not-nice thing here is that there is no automatic type generation, due to the different endpoint.


package.json
```diff
@@ -141,6 +143,7 @@
     "jeet": "*",
     "jquery": "*",
     "jquery.finger": "*",
+    "lodash": "^4.17.20",
```
dzucconi, line 146: I'd remove this and just use the native method

dzucconi: (Remember we don't have bundle splitting yet)

damassi: Second @dzucconi on that, but also note that https://github.com/lodash/babel-plugin-lodash will pluck things out if you just _have to_ lol

src/v2/apollo/index.tsx
```diff
@@ -31,6 +32,9 @@ const {
 
 const clientHttpLink = new BatchHttpLink({ uri: CLIENT_GRAPHQL_ENDPOINT })
 const serverHttpLink = new BatchHttpLink({ uri: GRAPHQL_ENDPOINT })
+const contentfulHttpLink = createHttpLink({
+  uri: 'http://localhost:5000/graphql/contentful',
```
dzucconi, line 36: Needs the ENV var?

broskoski: Yes, thank ye

damassi:
Funny how easy it can be... Curious how it works on the CMS side? Contentful calls a mutation to the GraphQL server which then reads in the data on the site?

broskoski:
@damassi contentful just offers a graphql api now (in addition to rest). they have their own CMS. Am I understanding your q?

damassi:
Ahh i see -- was thinking that it was a headless CMS that still required the user to host content themselves. Nice 👍

broskoski:
@damassi oh yea, I'm just making our own graphql endpoint to proxy through.

PR #2383, opened 2020-08-21
Fix top bar on error page

broskoski:
So apparently we already had a top bar on the error page, but when I refactored the topbar for client-side routing I broke it. This gets it rendering again, and makes search result items ordinary links when on the error page.


src/v2/components/TopBar/components/PrimarySearch/index.tsx
```diff
@@ -24,6 +25,7 @@ const Results = styled(Box)`
 interface PrimarySearchProps {
   scheme: 'DEFAULT' | 'GROUP'
   history: any
+  isErrorPage: boolean
```
dzucconi, line 28: This component shouldn't be remotely concerned with whether or not we're in an error page.

PR #2388, opened 2020-08-22
Rewrite confirm account pages

broskoski:
Fixes ARE-6

Screen Shot 2020-08-22 at 12 09 18 AM

Still to do:

- [x] Delete old confirmed app
- [x] Add ability to pass `email` to `ResendConfirmationEmail` mutation


broskoski:
@dzucconi mind taking a look when you get a sec?

PR #2389, opened 2020-08-23
Solves 401 error when session is bad

broskoski:
Fixes ARE-7

We previously had this handler on the root route, so putting it back (though still don't understand exactly how a session can go bad). This also corrects the logic on `GlobalNavHandler` for logout > redirect on a client-side unauthenticated error.

### DIFF
--- a/src/apps/app/App.ts
+++ b/src/apps/app/App.ts
@@ -47,5 +47,5 @@ App.get(
 )
   .get('/group/:id/invite/:code', ensureLoggedIn, ...resolve)
   .get('/tools/*', ensureLoggedIn, ...resolve)
-  .get('/', homePathMiddleware, apolloMiddleware, ...resolve)
+  .get('/', homePathMiddleware, ...resolve)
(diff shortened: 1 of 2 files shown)


src/apps/app/App.ts
```diff
@@ -20,7 +21,19 @@ const resolve = [
...
+        if (STATUS_CODE === 'UNAUTHORIZED' && req.url === '/') {
+          // This typically happens if the serialized user is "bad"
+          // or not actually logged in. If so: logout, then redirect somewhere.
+          // Falling through by using `next()` doesn't seem to actually purge the session.
+          req.logout()
+          return res.redirect('/log_in')
+        }
+
+        return next(err)
+      })
```
dzucconi, line 36: Not so sure this makes sense: wouldn't this log you out if you just happened to land on a private channel somehow?

broskoski: There’s a conditional there: 
`if req.url == '/' ‘

PR #2395, opened 2020-09-01
Trigger a full refresh when upgrading to premium to ensure cache is clear

broskoski:


src/v2/components/Billing/components/BillingForm/index.tsx
```diff
@@ -118,6 +119,13 @@ class BillingForm extends PureComponent<BillingFormProps, BillingFormState> {
       refetchQueries: [{ query: billingQuery }],
       awaitRefetchQueries: true,
     })
+      .then(() => {
+        return axios.get('/me/refresh')
+      })
+      .then(() => {
+        window.location.reload()
+        true
```
dzucconi, line 127: Why return `true`?

broskoski: Ah, copied pattern from user settings. Form component there asks for a return boolean value to indicate success.

PR #2403, opened 2020-09-11
Use picture tag to fallback to png when the browser doesn't support WEBP

broskoski:
The idea here is to fallback to our original stupid PNGs when the browser doesn't support WEBP AND when the image is an `Image` (which is the only case where it matters). 

Once all browsers are equal, this code shall be removed.


dzucconi:
The proxy should already handle this (content negotiation) and serve accordingly. It's not?

broskoski:
It does, this is just about falling back to transparent PNGs on Image blocks in Safari (since JPEGs flatten transparency wrong).

src/extension/src/components/ExtensionLogin/index.jsx
```diff
@@ -73,9 +73,9 @@ class Login extends Component {
               me.authentication_token
             )
           } catch {
-            // browser.storage.local.set({
-            //   authentication_token: me.authentication_token,
-            // })
+            browser.storage.local.set({
+              authentication_token: me.authentication_token,
+            })
```
dzucconi, line 78: What's this about?

package.json
```diff
@@ -184,6 +184,7 @@
     "react-split-pane": "^0.1.89",
     "react-stripe-elements": "^2.0.1",
     "react-transition-group": "^1.2.0",
+    "react-use-webp-support-check": "^1.0.5",
```
dzucconi, line 187: This is added but then never imported

src/v2/components/Cell/components/Konnectable/components/KonnectableGeneric/index.tsx
```diff
@@ -25,17 +25,24 @@ interface Props extends BoxProps {
...
   mode: Mode
+  fallbackToProxyJpeg?: boolean
 }
 
 export const KonnectableGeneric: React.FC<Props> = ({
   src,
   srcs,
   title = null,
   mode = Mode.RESTING,
+  fallbackToProxyJpeg,
```
dzucconi, line 36: Why a prop?

src/v2/components/Cell/components/Konnectable/components/KonnectableImg/index.tsx
```diff
@@ -12,7 +12,8 @@ const Container = styled(Box).attrs({
   width: 100%;
   height: 100%;
 
-  > img {
+  img,
+  picture {
```
dzucconi, line 16: `> img, > picture`

dzucconi:
Yeah I mean... this is a clever workaround, I just don't think it's warranted.

With the lack of `srcSet` it's objectively worse for those users: having flattened transparent pngs (which I still don't view as a 'bug') is obviously much worse than just having improperly sized images. This also presents problems with adopting the proxy elsewhere.

broskoski:
CLOSING

PR #2418, opened 2020-10-28
Add maxSize and note sizes on drop overlay

broskoski:
Responding to a fairly frequent question we get to help about uploading images or files to channel (about the maximum size, why certain images fail, etc). The ultimate idea would be to add custom validators to the drop zone (which apparently is possible: https://github.com/react-dropzone/react-dropzone/issues/321) which would communicate why a certain file was not valid.

I spent far too long trying to figure this out. The problem is that `acceptedFiles` is returned before a custom validator can can process the file (and there is no way to make those callbacks `async`). The only real option would be to overwrite `getFilesFromEvent`, but that is truly overkill. 

Maybe there is a better way to do this but right now I can't think of one.

Instead, just add the max file size for attachments, and add some help text to the drop zone overlay.

### DIFF
--- a/src/v2/components/UI/DropZoneUploader/index.tsx
+++ b/src/v2/components/UI/DropZoneUploader/index.tsx
@@ -76,6 +76,7 @@ const DropZoneUploader: React.FC<Props> = ({
     noClick: true,
     noKeyboard: true,
     accept,
+    maxSize: 250000000, // 25mb
   })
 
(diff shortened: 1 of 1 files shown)


broskoski:
https://linear.app/aredotna/issue/ARE-69/web-should-display-error-when-uploading-files-that-are-larger-than-the

PR #2423, opened 2020-10-30
Set width and height if image dimensions are available

broskoski:
Before:
Screen Shot 2020-10-30 at 3 23 58 PM

After:
Screen Shot 2020-10-30 at 3 24 00 PM


src/v2/components/Cell/components/Konnectable/components/KonnectableGeneric/index.tsx
```diff
@@ -25,18 +28,45 @@ interface Props extends BoxProps {
...
-  </Container>
-)
+}) => {
+  const { width: originalWidth, height: originalHeight } = originalDimensions
+
+  const { width, height } = scale({
+    width: originalWidth,
+    height: originalHeight,
+    maxWidth: Math.min(originalWidth, parseInt(constants.blockWidth)),
+    maxHeight: Math.min(originalHeight, parseInt(constants.blockWidth)),
```
dzucconi, line 51: You can/should set the maxWidth to just be the blockWidth here. You never want it to be larger than that presumably?

dzucconi: Oh I see, you don't want to scale it if it's smaller?

broskoski: Exactly

PR #2425, opened 2020-11-02
A11y

nel417:
Added titles to iframes to let users know what they are on. It helps if you know what you are going to hit "play" for. A video, audio, game, etc.

**Before**
Screen Shot 2020-11-02 at 4 07 30 PM

**After**
Screen Shot 2020-11-02 at 4 10 29 PM

**From** https://webaim.org/techniques/frames/
_When a screen reader user hears a list of frames, it is helpful to know the purpose of each one. Frame titles allow web developers to communicate the purpose of each frame to users of screen readers. The best titles for frames are brief and descriptive._

**Tested**
**OS**:  macOS Mojave 10.14.6
**Browsers**: Google Chrome, Mozilla Firefox (most recent)
**Screen Reader** Apple VoiceOver

### DIFF
--- a/src/apps/about/templates/index.jade
+++ b/src/apps/about/templates/index.jade
@@ -23,6 +23,7 @@ block container
             src='https://player.vimeo.com/video/227975322?autoplay=0&autopause=0'
             frameborder='0'
             allowfullscreen
+            title='about video'
           )
       br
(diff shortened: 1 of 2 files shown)


broskoski:
@nel417 thank you so much!!

PR #2436, opened 2021-01-23
Refactor BillingForm and PlanChanges to allow subscriptions without a source if the balance is 0

broskoski:
The front-end version of https://github.com/aredotna/api/pull/1282. Also refactors BillingForm and PlanChanges into functional components with hooks.


src/v2/hooks/useMergeState/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+        })
+      } else {
+        setState(state => ({ ...state, ...update }))
+      }
+    },
+    [setState]
+  )
+
+  return [state, updater]
+}
```
dzucconi, line 49: ⬆️ ❓

broskoski: Came from https://github.com/react-restart/hooks/blob/master/src/useMergeState.ts. I was afraid of taking too far of a departure from your original code (which used `setState`).

dzucconi: I don't see it being used.

dzucconi: Oh, github had a bunch of diffs collapsed on me

src/v2/components/Billing/components/BillingForm/index.tsx
```diff
@@ -30,93 +27,172 @@ import PlanChanges from 'v2/components/Billing/components/PlanChanges'
...
+}) => {
+  const { customer } = me
 
-  static defaultProps = {
-    plan_id: null,
-    onSuccess: () => null,
-  }
-
-  state = {
+  const [state, setState] = useMergeState<BillingFormState>({
```
broskoski, line 82: @dzucconi ⬆

src/v2/hooks/useMergeState/index.tsx
dzucconi: I see this as far more confusing than just writing a reducer for the component specific state. Particularly since it messes with what you'd expect `setState` to be doing.

dzucconi: Also this is basically just

src/v2/components/Billing/components/BillingForm/index.tsx
```diff
@@ -30,93 +27,172 @@ import PlanChanges from 'v2/components/Billing/components/PlanChanges'
...
+import {
+  SubscribeToPremiumWithoutToken,
+  SubscribeToPremiumWithoutTokenVariables,
+} from '__generated__/SubscribeToPremiumWithoutToken'
+
+type OperationsEnum =
+  | 'CHANGE_PLAN_ID'
+  | 'CANCEL_PREMIUM_SUBSCRIPTION'
+  | 'DOWNGRADE_TO_LIFETIME'
+  | 'APPLY_COUPON_CODE'
```
dzucconi, line 52: 👍

src/v2/components/Billing/components/BillingForm/index.tsx
```diff
@@ -30,93 +27,172 @@ import PlanChanges from 'v2/components/Billing/components/PlanChanges'
...
-  coupon_code: string
+  mode:
+    | 'resting'
+    | 'error'
+    | 'processing'
+    | 'canceled'
+    | 'subscribed'
+    | 'card_changed'
+  operations: OperationsEnum[]
+  planId?: SupportedPlanEnum | 'basic' | 'lifetime'
```
dzucconi, line 69: This seems weird? Shouldn't these be encapsulated in that enum as well?

broskoski: Yeah it's a bit weird, but those are not supported plans (but they are values that could potentially be set).

PR #2439, opened 2021-02-01
Address banner bugs

broskoski:


broskoski:
Fixes https://linear.app/aredotna/issue/ARE-120/banner-covers-bottom-button-on-billing-page

PR #2442, opened 2021-02-25
Extract image from right-click location in browser extension

broskoski:


src/lib/PaneDataExtractor.ts
```diff
@@ -63,14 +64,91 @@ class PaneDataExtractor {
...
+    })
+
+    return stack
+  }
+
   extractImageFromHTML = data => {
     const html = new DOMParser().parseFromString(data['text/html'], 'text/html')
       .body.firstChild
+
+    console.log({ html })
```
dzucconi, line 150: 🧇

broskoski:
instagram-extension

PR #2447, opened 2021-03-24
Update roadmap and perks page

broskoski:


src/v2/styles/constants.ts
```diff
@@ -65,11 +69,31 @@ export const RADII = {
...
+  blockPreviewWidth: string
+  legacyUnit: string
+  topBarHeight: string
+  // TODO give these real types
+  media: any
+  radii: any
+  breakpoints: any
+  space: any
+  z: any
+}
```
dzucconi, line 90: You can just declare the theme in a const then `as const` and use `type Theme = typeof THEME` or whatever

PR #2471, opened 2021-05-03
Bare bones implementation of table view

broskoski:
It's very likely most of this will change, just getting this down as a start to iterate from. This route will only mount in the dev env.

Screen Shot 2021-05-03 at 6 08 22 PM


broskoski:
Gonna go ahead and merge here and keep going

PR #2482, opened 2021-05-27
ARE-162: Apollo client v2 -> v3

tvler:
This PR upgrades Apollo client from v2 to v3.

Biggest changes:

- Cache system has been reworked to not normalize types that should only exist when embedded in a parent model (UserCounts, ChannelCounts, etc), now that apollo has removed the "magic normalization" of v2.
- Getting apollo mocks working with storybook again will be a fast-follow.

There are _theoretical_ performance improvements with this upgrade, but we'll have to see if things are actually feeling faster on prod! This change mostly positions us to be in a good place for setting up cache TypePolicies for more complex queries going forward.

The biggest reason this diff is so huge is because the upgraded apollo package no longer minifies the auto-generated schema.json file (which I personally like, just for readability). Not factoring that in, the LOC is pretty stable if not down a bit.

Further info on Apollo client v3 from the client architect: https://www.youtube.com/watch?v=n_j8QckQN5I


broskoski:
Hell yes @tvler! So exciting.

Have been looking at the diff already, so gonna go ahead and merge this now so we can try it out on staging.

Thank you and congrats on a gigantic first PR :)

PR #2520, opened 2021-08-09
ARE-78: Refactor channel grid loading

tvler:
# Description

This PR moves all network IO business logic out from the ChannelContents component and into a new hook called [usePaginatedBlocks](https://github.com/aredotna/ervell/pull/2520/files#diff-acc09e6cc6777bbd32a75c3b9d8c4da941d4389ae9167411fa5ca0c8fa0d0495).

usePaginatedBlocks has an API that looks like this:

```ts
type UsePaginatedBlocksApi = {
  blocks: ChannelBlokksPaginated_channel_blokks[]
  contentCount: number
  getPage: (pageNumber: number) => void
  hasQueriedPage: (pageNumber: number) => boolean
  getPageFromIndex: (index: number) => number
  removeBlock: (args: { id: number; type: string }) => void
  moveBlock: (args: { oldIndex: number; newIndex: number }) => void
  addBlock: () => void
}
```

**This will allow us to build out completely new UIs for channels, now that the data business logic is completely abstracted from UI.** Also, we could potentially abstract this hook even more to work with _any_ paginated field regardless of its type (followers, following, etc)

To get a better feeling of this, here's the line diff of the ChannelContents component, down by about 50%:

Screen Shot 2021-08-09 at 10 11 22 AM

Under the hood, all internal state is being handled by apollo cache, which will help us down the road now that we are storing block data in a central store that other features and components could potentially pull from in the future.

Other changes:

- Updated typescript
- Updated apollo client
- Updated eslint-plugin-react-hooks
- Updated react-sortable-hoc
- Improved the paginationWithPageAndPer apollo cache field policy
- types the withIsSpiderRequesting HOC
- improves the usePusher type

# Testing

- Websockets are working as expected
- Adding/removing/moving blocks works as expected
- Scrolling thru big channels works as expected


src/extension/safari-extension/background.ts
```diff
@@ -47,7 +47,7 @@ export class SafariApp {
         })
       )
       if (resolveNow) {
-        resolve()
+        resolve(undefined)
```
tvler, line 50: This was a new type error that the upgraded typescript found

src/v2/components/ChannelContents/components/ChannelContentsItem/index.tsx
```diff
@@ -24,64 +64,114 @@ interface Props {
...
+      (newIndex: number) => {
+        onChangePosition({
+          oldIndex: index,
+          newIndex,
+        })
+      },
       [index, onChangePosition]
     )
 
+    const intersectionObserverCallback = useCallback<
```
tvler, line 103: There's a pretty cool optimization we can make later, where we only call `onItemIntersected` if the element is tracked as being in view for a duration of time rather than immediately.

The end result will be that a block's page will only get queried if a block is in view for ~0.2 seconds instead of immediately. This will reduce the amount of queries in the event that a user is quickly scrolling down to the bottom of a grid, because we won't be loading the intermediate pages since they would only briefly be in view.

src/v2/components/ChannelContents/index.tsx
```diff
@@ -1,124 +1,87 @@
...
-      connectable => {
-        loadSkeleton({
-          client,
-          channelId: channel.id,
-          pageSkeleton: [connectable],
-          collection,
-          queryOptions: {
-            fetchPolicy: 'network-only',
+      ({ id }: { id: string | number }) => {
+        client.query<ConnectableBlokk, ConnectableBlokkVariables>({
```
tvler, line 68: Because channel content is stored in apollo cache, all we need to do to revalidate a block is to query for that same block again! Then apollo will automatically update the cached data _and_ re render any active queries that are referencing the value

src/v2/components/ChannelContents/lib/usePaginatedBlocks.ts
```diff
@@ -0,0 +1,342 @@
...
+              insert_at: prevCount - newIndex,
+            },
+          })
+          .then(() => {
+            // After the mutation we need to revalidate pages that have
+            // already been queried, if a block moves into the page
+            // that previously was in a page that wasn't loaded.
+            // This is because we try to not call getPage if hasQueriedPage
+            // returns true, and if an unloaded block moves into the page
+            // then there will be no opportunity for that block to ever load.
```
tvler, line 283: Just pushed up a bugfix I found when moving blocks. If you're in a channel with more than 1 page of blocks and press the "Send to bottom" button on the 1st block, you can get into a state where a block that hasn't been downloaded yet is moved into a page that we are tracking as "already downloaded". The end result is that block is permanently in a loading state until refresh.

This fixes is by querying for all pages in between the block's old position and new position that need to be revalidated.

tvler:
This commit that I just pushed up (https://github.com/aredotna/ervell/pull/2520/commits/538673125ece4d4993921f2f55f985626b5f3e26) alters the networking IO so that the entire layout of the grid can get rendered in SSR. 

This means that the page's first initial render in the browser will have the same document height as it does when a channel is fully loaded. Which means that _scroll position_ will be maintained on channel pages when navigating in your browser!

tvler:
Found a few more bugs, so hold off on merging for now

tvler:
Cool, remaining bugs should be fixed! Here's the diff of the fixes https://github.com/aredotna/ervell/pull/2520/files/538673125ece4d4993921f2f55f985626b5f3e26..1cdffb5862ddf1f8031140c19a186f2c7ed64c60

1. Fixes the "updateConnectable" pusher callback
2. Fixes the "addBlock" callback so that the block cache gets completely deleted and revalidated
3. Removes all the arguments from the addBlock callback in child components now that "addBlock" is basically just a "trigger" to completely revalidate a channel and doesn't really care about the particular block that was added

PR #2533, opened 2021-08-30
lib: Better error handling from express-reloadable

damassi:
Caught an issue last week where `express-reloadable` was [silently consuming errors](https://github.com/artsy/express-reloadable/pull/45) lower than the reloadable mount point, so wanted to follow up with a version bump over here.


broskoski:
Thank you @damassi !!!

PR #2534, opened 2021-08-31
Refactor RSS app

broskoski:


src/apps/rss/index.ts
```diff
@@ -0,0 +1,71 @@
...
+            return next(err)
+          })
+      }
+    })
+  },
+]
+
+app.get('/explore/feed/rss', ...resolve)
+app.get('/:username/feed/rss', ...resolve)
+app.get('/:username/:slug/feed/rss', ...resolve)
```
dzucconi, line 66: If this is using react-router; why bother spinning up a new app for it?

dzucconi: Oh I see; custom layout; yeah.

broskoski: the specific resolve and caching

src/v2/pages/rss/ChannelRss/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+import {
+  ChannelRss as ChannelRssType,
+  ChannelRssVariables,
+} from '__generated__/ChannelRss'
+import { RssBlock } from '../components/RssBlock'
+import RssLayout from '../components/RssLayout'
+import channelRssQuery from './queries/channelRss'
+
+interface ChannelRssProps {
+  params: any
```
dzucconi, line 12: Shouldn't this just be `params: { slug?: string }`?

dzucconi (approved):
Cool

PR #2554, opened 2021-09-21
Fix pusher updating for new blocks transitioning to processed blocks

tvler:
This PR potentially fixes a bug where a PendingBlock will never update to the final block.

While investigating the issue I came up with 2 potential issues:

1. A new pusher client is initialized every time a channel page is updated. If you go to a channel and add 4 blocks to it, you can see in the network tab that multiple pusher connections are established:

Screen Shot 2021-09-20 at 3 55 23 PM

This creates the possibility of a window of time, after the first channel disconnects and before the new one connects, where an update event can come in and not get captured.

2. Query deduplication is on for usePaginatedBlocks calls. Query deduplication is a heuristic that apollo implements which tries to cancel queries if the same query is currently in-flight. This means that if `getPage(1)` was called a few seconds ago, another call to `getPage(1)` would get pointed to the first (potentially out of date) query instead of making a new network call.

It's kind of a bummer that this issue is flakey but I'm feeling pretty confident that it's happening because of one or a combination of these reasons. If my changes don't solve the issue we can do a more aggressive fix but it will really increase the complexity of the usePaginatedBlocks callback, which I'd rather not do unless totally necessary. These changes I'm introducing are beneficial regardless of the bug or not though, so not all would be lost if this doesn't work out :)

### DIFF
--- a/src/v2/components/ChannelContents/index.tsx
+++ b/src/v2/components/ChannelContents/index.tsx
@@ -1,4 +1,4 @@
-import React, { memo, useEffect, useCallback } from 'react'
+import React, { memo, useCallback } from 'react'
 import { SortableContainer } from 'react-sortable-hoc'
 
 import { ChannelContents as ChannelContentsData } from '__generated__/ChannelContents'
@@ -20,8 +20,6 @@ const SortableGrid = SortableContainer(({ onSortEnd: _onSortEnd, ...rest }) => (
 
 interface Props {
   channel: ChannelContentsData
-  pusherChannel?: any
-  socket?: any
 }
 
 interface ExtendedProps extends Props {
@@ -51,7 +49,7 @@ const parsePayload = (payload: any): PusherPayload => {
 }
 
 const ChannelContents: React.FC<Props> = WithIsSpiderRequesting<ExtendedProps>(
-  memo(({ channel, pusherChannel, socket, isSpiderRequesting, ...rest }) => {
+  memo(({ channel, isSpiderRequesting, ...rest }) => {
     const {
       blocks,
       getPage,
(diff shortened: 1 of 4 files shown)


src/v2/pages/channel/components/ChannelContentsWithData/index.tsx
```diff
@@ -49,25 +38,8 @@ export const ChannelContentsWithData: React.FC<Props> = WithIsSpiderRequesting<
         const { channel: clientChannel } = data
         const channel = { ...serverChannel, ...clientChannel }
 
-        const shouldSubscribeToPusher =
-          !isSpiderRequesting && channel.can.add_to
-
-        if (shouldSubscribeToPusher) {
```
tvler, line 55: Perviously we were calling `setupPusherChannel` every time this component gets rendered, which creates new `pusherChannel ` and `socket` values that don't have reference equality across renders. This means that we were unnecessarily rendering `ChannelContents` a ton more than needed (ChannelContents is a pure component)

src/v2/components/ChannelContents/index.tsx
```diff
@@ -106,21 +104,13 @@ const ChannelContents: React.FC<Props> = WithIsSpiderRequesting<ExtendedProps>(
...
     usePusher({
-      channel: pusherChannel,
+      channelId: channel.id,
+      skip: !isSpiderRequesting,
       onCreated: addBlock,
       onUpdated: updateConnectable,
       parsePayload: parsePayload,
     })
 
-    useEffect(() => {
```
tvler, line 115: All the logic to unsubscribe from a channel is now encapsulated in `usePusher` itself!

src/v2/components/ChannelContents/index.tsx
```diff
@@ -106,21 +104,13 @@ const ChannelContents: React.FC<Props> = WithIsSpiderRequesting<ExtendedProps>(
     )
 
     usePusher({
-      channel: pusherChannel,
+      channelId: channel.id,
+      skip: !isSpiderRequesting,
```
broskoski, line 108: And this should actually not be called if the person cannot add to the channel as well:

```js
const shouldSubscribeToPusher =
           !isSpiderRequesting && channel.can.add_to
```

broskoski (changes requested):
Looks good otherwise!

src/v2/components/ChannelContents/index.tsx
tvler: fixed and simplified the boolean logic here https://github.com/aredotna/ervell/pull/2554/commits/9ac0b5eb0700f3cb859f6bec31525d7d38438f9c

broskoski (approved):

PR #2556, opened 2021-09-21
[WIP] Remove revalidation from moveBlock function

broskoski:
@tvler looks like this is where the issue is coming from. Not exactly sure what's going on, this does make things marginally better but obviously stops covering the case when a page isn't loaded.


broskoski:
Issue here: https://linear.app/aredotna/issue/ARE-206/sorting-blocks-causes-blocks-to-refresh-and-sorted-position-is-not

tvler:
Down for this solution 👍 will investigate the underlying issues

PR #2560, opened 2021-09-23
Make sure the block doesn't already exist before calling addBlock

broskoski:
@tvler can you take a quick look if you have a second?


broskoski:
Gonna merge to check it on staging!

PR #2576, opened 2021-10-17
ARE-151: Table pagination part 1

tvler:
_Note: The code diff automatically minimizes [usePaginatedBlocks.ts](https://github.com/aredotna/ervell/pull/2576/files#diff-d964da62bf5c895e11517bd84df2e2b424b1e26c7d29802ab043da585d834f8c) but please expand it when reviewing! That's where the bulk of this code change is :)_

This PR upgrades usePaginatedBlocks so that the hook can work in different contexts. Particularly:

1. With different graphql channel queries
2. With different graphql block queries

These updates make it possible to swap out the useQuery call in the table component with usePaginatedBlocks! In the next PR I'll be working on adding pagination beyond 25 blocks. For now I just wanted to ship the initial updates to usePaginatedBlocks since it's at a pretty good landing point.

Changes to usePaginatedBlocks:

- The channel graphql query is now a required argument, rather than internally using channelBlokksPaginatedQuery. This way, the table component can use its own channel query that's completely different from the grid query. To achieve this, we define a minimum shape required for internal use of the query, then we assert that the passed-in generic types extend that.
- The "blocks per page" value is now a required argument since we want to load more or less blocks per page in different contexts
- The individual block graphql query is now an argument, rather than internally using ConnectableBlockQuery. If passed in, this unlocks the ability to use the updateBlocks method through [function overloading](https://www.typescriptlang.org/docs/handbook/functions.html#overloads) in typescript. This had to be changed because we now aren't sure that the individual block query's shape will match the block shape of the channel query.
- Adds a new internal helper function getQueryFromCache. This function returns the currently cached query data and is useful if you want to use this data in a memoized function without re-memoizing every time the query data changes (which happens a lot). This internal helper function is now used in the getBlocksFromCache and updateBlocks function instead of re-implementing this functionality in multiple places.
- Moves usePaginatedBlocks to v2/hooks/usePaginatedBlocks and getConnectableType to v2/util/getConnectableType since they're now used in different places around the codebase

Other code changes:

- Removes unused graphql generated types
- Changes ChannelTablePage from using query components to query hooks for performance improvements

QA:

- Make sure the grid view is operating as expected
- Make sure the table view is operating as expected (no changes to that yet)

### DIFF
--- a/src/v2/hooks/usePaginatedBlocks/index.ts
+++ b/src/v2/hooks/usePaginatedBlocks/index.ts
@@ -13,7 +13,7 @@ import moveConnectableMutation from 'v2/components/ChannelContents/mutations/mov
 import { getConnectableType } from 'v2/util/getConnectableType'
 
 /**
- * The required minimum shape for the channel query
+ * The minimum required shape for the channel query
  */
 type RequiredChannelQueryData = {
   channel: null | {
@@ -33,7 +33,7 @@ type RequiredChannelQueryData = {
 }
 
 /**
- * The required minimum shape for the channel query variables
+ * The minimum required shape for the channel query variables
  */
 type RequiredChannelQueryVariables = {
   id: string
@@ -42,7 +42,7 @@ type RequiredChannelQueryVariables = {
 }
 
 /**
- * The required minimum shape for the block query
+ * The minimum required shape for the block query
  */
 type RequiredBlockQueryData = {
   blokk: null | {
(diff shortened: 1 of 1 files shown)


src/v2/components/ChannelContents/index.tsx
```diff
@@ -145,32 +163,30 @@ const ChannelContents: React.FC<Props> = WithIsSpiderRequesting<ExtendedProps>(
     }
 
     return (
-      <>
```
tvler, line 148: This fragment isn't needed

PR #2604, opened 2021-11-20
Table pagination

tvler:
This PR adds full pagination support for the table UI, made possible by the usePaginatedBlocks hook!

# Networking logic

The networking logic is basically the same as how the grid view works. Every block now has an intersection observer attached to it. When a block is scrolled into view, we check if the page has already been queried for. If it hasn't, then we call getPage for the page that the block is in.

The table renders a row for every connectable item the channel has. It doesn't matter if an item has been loaded/not loaded. This means that there is a very small amount of layout shifts that happen while scrolling around a table which improves UX and browsing performance.

# react-table improvements

All of the arguments/values associated with react-table are now properly typed. Along with that, all arguments for react-table are now properly memoized. This should reduce the number of times react-table needs to rerender.

Sorting isn't handled by react table anymore. This is because react-table's sorting functionality only would properly work if every block has been loaded (it wouldn't know where to reposition null blocks that haven't been loaded in yet). All sorting callbacks now directly update the channel query variables passed into usePaginatedBlocks.

# usePaginatedBlocks updates

The biggest change for usePaginatedBlocks is that we now allow the channel query variables to change, which allows things like sorting to work.

Most of this PR was worked on with the strictNullChecks tsconfig value turned on (I was running into a lot issues upgrading our table rows to be able to render null values and this setting really helped improved the debugging process). As an effect of that setting being on, I made some null handling improvements to this hook since it's very important that can correctly handle null/non-null values.


src/v2/components/ChannelContents/index.tsx
```diff
@@ -131,7 +131,7 @@ const ChannelContents: React.FC<Props> = WithIsSpiderRequesting<ExtendedProps>(
 
     usePusher({
       channelId: channel.id,
-      shouldSubscribe: !isSpiderRequesting && channel.can.add_to,
+      shouldSubscribe: !isSpiderRequesting && !!channel?.can?.add_to,
```
tvler, line 134: A lot of these null-checking changes are because I had strictNullChecks on in my local branch while working on the table. I didn't want to go overboard with upgrading our entire codebase, but I made some changes to files that I consistently had open to increase clarity.

src/v2/components/ChannelTableContents/components/ChannelRow/index.tsx
```diff
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { forwardRef } from 'react'
```
tvler, line 1: Every table row needs to be a forwardRef component so that we can attach an IntersectionObserver to it

src/v2/components/ChannelTableContents/components/ContentCell/index.tsx
```diff
@@ -34,8 +34,12 @@ const Source = styled(Text).attrs({
 export const ContentCell = ({
   value: content,
 }: {
-  value: ChannelTableContentsSet_channel_blokks
-}): JSX.Element => {
+  value: TableData
+}): JSX.Element | null => {
+  if ('isNull' in content) {
```
tvler, line 39: We need to check if a table row is null now

broskoski:
Looks and works awesome, and I've checked it out locally. Well done sire 💯

PR #2614, opened 2021-12-11
[WIP] Type and connector filter

broskoski:


src/v2/hooks/usePaginatedBlocks/index.ts
```diff
@@ -351,16 +360,23 @@ export function usePaginatedBlocks<
...
    */
+
+  const { data, loading, error } = useQuery<
+    ChannelContentCount,
+    ChannelContentCountVariables
+  >(CHANNEL_CONTENT_COUNT, {
+    fetchPolicy: 'cache-and-network',
+    variables: {
+      id: channelId,
+      type: type,
```
broskoski, line 371: @tvler this PR isn't done yet, but heads up – passing `type` and `user_id` to the count now (to get an updated count when a filter is applied). Also had to change the `fetchPolicy` here to make it work. Does this make sense?

tvler: Yeah interesting. I _think_ that if this stays as cache-only, then the count data will eventually correctly populate once a filter is applied. But there will a period of time right after the user applies a filter where apollo cache obviously won't have that data. Eventually though, the passed-in channel query will return a fresh counts value that this cache-only query can properly read from.

If this isn't what you were experiencing when building this feature then lmk and I can look into it.... but I think that cache-only _should_ work!

PR #2618, opened 2021-12-18
ARE-154: Table drag and drop

tvler:
This PR adds drag + drop to the table view. This is done using react-sortable-hoc, which is already used in the app for grid views.

Other changes:

- Makes the default sort value for table views "Sorts.Position" instead of "Sorts.UPDATED_AT"
- Only allows drag+dropping functionality if the current sort is Sorts.Position (wouldn't make any sense otherwise)
- Updates the usePaginatedBlocks "updateCache" function to use "updateQuery" internally instead of  "client.cache.modify." https://www.apollographql.com/docs/react/caching/cache-interaction/#using-updatequery-and-updatefragment This is because client.cache.modify wasn't able to perform a cache update on a specific cache instance of a channel's "blokks." When you re-sort a channel, you generate multiple "blokks" instances on a single channel's cache, and the previous implementation was updating all of them (not just the "Sorts.Position" blokks instance). This new method fixes this bug because the API allows you to pass in the query variables, and apollo is able to determine which blokks instance that particular set of query variables it's tied to.

Todo after this PR

- Fine-tune the UI. Currently the drag+drop UI is just a gray square. We can build out the "Settings" column & corresponding buttons in a separate PR


src/v2/components/ChannelTableContents/lib/constants.ts
```diff
@@ -1 +1,13 @@
...
+import { ColumnIds } from './types'
+
 export const FIRST_COLUMN_WIDTH = `35%`
+
+/**
+ * If we want to allow a column to be sortable,
+ * add the column's ID to this object with the Sorts
+ * value it should control
+ */
+export const columnIdsToSorts: { [key in ColumnIds]?: Sorts } = {
```
tvler, line 11: Moved this here because it was causing a circular dependency

PR #2619, opened 2021-12-21
Fix table view rendering

broskoski:
@tvler `ssr: false` fixed the rendering issue :)


src/v2/components/ChannelTableContents/components/PotentiallyEditableBlockCell/index.tsx
```diff
@@ -99,7 +100,10 @@ const PotentiallyEditableBlockCellNonNull = ({
   const [verifyEditable, { data, loading }] = useLazyQuery<
     VerifyEditableBlock,
     VerifyEditableBlockVariables
-  >(verifyEditableQuery, { variables: { id: block.id?.toString() } })
+  >(verifyEditableQuery, {
+    variables: { id: block.id?.toString() },
+    ssr: false,
```
broskoski, line 105: @tvler here

tvler: Nice! Very weird bug…

PR #2908, opened 2023-03-07
Remove ability checking on initial load and wait until context menu is activated

broskoski:
Almost all of this is in [src/v2/components/ConnectableContextMenu/index.tsx](https://github.com/aredotna/ervell/compare/improve-channel-loading?expand=1#diff-5a66b7fc26e02eaf72eded2c31582cdd3c21f09388a0d529d1f5be1fe388baeb) the rest is basically just fixing the types. No need for a thorough review, just showing you in case you're curious

### DIFF
--- a/src/v2/components/ConnectableContextMenu/index.tsx
+++ b/src/v2/components/ConnectableContextMenu/index.tsx
@@ -1,5 +1,4 @@
 import React from 'react'
-import styled from 'styled-components'
 
 import { ConnectableContextMenuChannel as ConnectableContextMenuChannelData } from '__generated__/ConnectableContextMenuChannel'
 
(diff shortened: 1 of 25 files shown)


src/v2/components/ConnectableContextMenu/index.tsx
```diff
@@ -24,6 +39,116 @@ export const ConnectableContextMenu: React.FC<Props & BoxProps> = ({
...
+  const { data, loading, error } = useQuery<
+    ChannelContextMenuType,
+    ChannelContextMenuVariables
+  >(channelContextMenu, {
+    variables: {
+      id,
+    },
+  })
+
+  if (loading) return <></>
```
dzucconi, line 128: Why not return a loading state here; this is going to feel really weird otherwise.

PR #2956, opened 2023-11-12
Removed upcoming price changes

jondashkyle:
This was introduced at the beginning of the year, and is no longer relevant.


broskoski:
Thank you!

